PR #2, opened 2021-07-01
Delete README.md

patudom:
Removing placeholder readme file, so Nick's updated version is displayed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,1 +1,0 @@
-# cosmic-ds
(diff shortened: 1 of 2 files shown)


nmearl:
This is great and definitely would fix the issue. I think, though, we might want to go with the more general solution of enabled package data on install included in #3. In that case, you can simply revert your change to the `setup.cfg` file, and we can merge this in for the `README.md` removal.

PR #7, opened 2021-07-07
Add histogram view of student age measurements

patudom:
-Add data file that was output from @Carifio24's simulation script. (Ignore others for now)
-Add age_distr_viewer functionality to app.py and app.vue

Still buggy. @nmearl, the histogram plot does not appear unless I pull up the browser console. Only after I open and close the console, the graph appears. Any ideas what could be causing this behavior?


patudom:
@heywooddogwood: My 2nd commit here is meant to help you and me figure out what needs to go where in the UX code.

@nmearl: Can you please verify that I identified the different pieces of code correctly? Thanks!

patudom:
@Carifio24 & @nmearl, regarding the histogram plot not showing up until after I open & close the console, that is the behavior I'm seeing in Firefox & Safari. If I run the notebook in Chrome, pulling up the console does not make the histogram appear. Chrome is also the only browser that brings up the console as a separate window.

nmearl:
The problem is that when the page initially loads, all the viewers load. Since the histrogram isn't visible at the time, the container of the viewer is considered to have zero height/width. The figure loads at these dimensions, and it takes a window resize event to trigger a relayout (i.e. you should also be able to trigger the viewer to show just by resizing your browser window in addition to calling up the console).

To get around this, you can have the viewer load only when it's actually visible on the page. To do it, just wrap the in a `v-lazy` component like this:

```html
<v-lazy>
  <jupyter-widget
    style="height: 300px"
    :widget="viewers.age_distr_viewer"
  >
  </jupyter-widget>
</v-lazy>
```

patudom:
@nmearl Aha! Thanks so much for helping to sort this out.

heywooddogwood (approved):
Everything looks good, although I'm not sure why there are now four paginations under Estimate Galaxy Distance, given that there are still only two tabs. I plan to remove them on my next pull request, unless there is a reason for the addition.

patudom:
@heywooddogwood Thanks, and sorry about the 4 paginations. I made a random change to verify that Nick's version to reload the app.vue file works, and I forgot to change it back.

PR #9, opened 2021-07-08
Add simple reload functionality

nmearl:
This PR adds a `reload` method to the `Application` object that would allow easily reloading the UI without having to restart the python kernel.


patudom:
Thanks for this, @nmearl.  FYI, this didn't work for me off the bat. I just talked to @Carifio24, and he suggested that I run the installation in editable mode (pip install -e .) and that allowed the notebook to reload the app.vue file.  

Should I add a note about that to the README instructions?

cc'ing @heywooddogwood in case you end up having the same issue I did.

PR #12, opened 2021-07-20
CSS Example Notebook

nmearl:
Include is a css example notebook with some quick details on how to edit the css through the python interface. This does not include (yet) an example of how to edit the css using a style sheet.

Screen Shot 2021-07-20 at 00 42 07

### DIFF
--- a/notebooks/Example Plot Styling.ipynb
+++ b/notebooks/Example Plot Styling.ipynb
@@ -1,0 +1,116 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 70,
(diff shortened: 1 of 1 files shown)


patudom:
Thanks, @nmearl, this is very helpful. I'm trying to edit your sample notebook to specify that tick_format should include thousands comma separators ({:,.0f}), and it's not working. Any idea if this is possible, and if so, what the magic syntax is?  Thanks!

nmearl:
Yep, don't use the curly braces or the colon, so simply `',f'` will do. I've updated the notebook with an example.

PR #15, opened 2021-07-26
Dialog component example

nmearl:
This PR includes an example of creating a dialog component, instances it in the `app.py` file, and referencing it in the `app.vue`. Note that there are _many_ ways to handle dialogs, and it may be easiest to simply include the Vue code directly without making a separate component.


patudom:
Merged these commits to #18. Closing this branch.

PR #19, opened 2021-08-03
Spectrum js code from SED Spectrum Lab project

patudom:
These files are a stripped down version of Alaalden Ibrahim's code at https://waps.cfa.harvard.edu/microobservatory/spectrum/playground.html, used here with permission from Alaalden & Spectrum Lab PI Mary Dussault.

To view:
- Load the html file
- Click "choose data file to upload txt"
- Select the included sample data file (sun_ref_colina96.txt)
- Click "Display file"

Ultimately, the "display file" functionality will be tied to the students' galaxy selection in WWT.

The Spectrum Lab has a lot more functionality than we need. I tried to strip away most of what we aren't using, but I did not do this in a very efficient way, so this should be cleaned up further.


patudom:
@nmearl This is the js to display the galaxy spectra that we will use in the velocity measurement tab. Can you show us how to wrangle this into the ipyvuetify app, please? Thanks!

patudom:
We think it will be more straightforward to write new python code with the spectrum tool functionality instead of trying to wrap in this additional layer of .js into the ipyvuetify app, so I am closing this for now and will open a new issue to serve as the thread for the python version of this functionality.

PR #21, opened 2021-08-05
Add wtml file from WWTA Hubble app

patudom:
This is the WTML file WWT uses to display the overlaid galaxy images available in the http://projects.wwtambassadors.org/galaxy-history-universe/ app.


Carifio24:
Cool! I'll work on getting this displayed in the WWT viewer in the app.

PR #22, opened 2021-08-05
Add initial line-fitting functionality

Carifio24:
This PR adds the basic functionality for adding the best-fit line to the student's data via a button press on the second screen. Each time that `vue_fit_lines` is called for a viewer, any previous lines from data/subsets in that viewer are removed. Since the glupyter viewer doesn't have an option to draw lines, my workaround was to add a `LinesGL` mark onto the viewer's figure image. I also added a second scatter viewer, so that the fit line doesn't appear on the first screen.

@nmearl Do you think this seems like a reasonable approach, relative to the app architecture as a whole? If so, I can add the same functionality for student/class/all public data on the third screen.


nmearl:
The approach here certainly works, but I would consider the case where you e.g. want all or a subset of students, but filtered by galaxy type or class id. In that case, you'd have to have preemptively created subsets for each combination or end up having to combine data across different layers, which could be tedious. 

This is sort of the case I was considering when doing the dynamic subset creation. In my example, I had imagined you might (1) let the user select their criteria (galaxy type + class 1 + students 1-10), then (2) create the subset and convert it to a layer on-the-fly, and finally (3) fit a line to the new layer. Once done, the layer can be disposed of. This would allow us to keep all the data together and simply "filter" when needed.

Of course, I haven't thought of all use cases/corner cases, so this definitely open to discussion/improvement.

Carifio24:
I was thinking that if we wanted to (for example) automatically fit and plot a new line when the student changes their selection, we could call `vue_fit_lines` at the end of `vue_galaxy_selected` and `vue_class_selected`, or whatever their analogues are in the app itself. The old subset gets deleted, the line-fitting function removes the line for the old subset, and fits one for the new subset.

I guess the real issue there is that, as written, `vue_fit_lines` fits a line to every visible layer, which we may not want in a case like your notebook example. Some flexibility with what we actually fit to would be nice for that sort of scenario.

patudom:
@nmearl, see my comment on #23. I think allowing selections on 2 independent axes could end up being very confusing for students, so we will likely want to limit the screens to allow only 1 type of selection at a time.  

Regarding @Carifio24's last comment, I think students may get confused if the fit lines are automatically updated every time they change their selection. My suggestion is that the lines should all disappear whenever students make a new selection, and they have to click a button to display the updated fit line.

Another question that @Carifio24 and I discussed is whether we should show 1 line that is a fit to all displayed data, or if there should be separate fit lines displayed for each subset of data. It's possible (likely?) that offering too many choices like this will result in cognitive overload for students, but I could imagine keeping both buttons around while we are testing and then we can decide where we want to use each type of fit, or whether there is a way to sensibly offer both choices:
- Fit one line to all data currently selected/displayed
- Fit separate lines to each subset of data

Carifio24:
So I've added a couple more options to the line fitting. The `vue_fit_lines` function now has options to

- Specify which layers to include in any fitting (this is done via the UUID of the data object, as we can easily grab the data objects via the app's data collection). The default is to just use all visible layers.
- Indicate whether or not we want to aggregate the data in these layers for a line fit, or do a line fit for each layer
- Specify what we want to do with already-present lines (i.e. keep or delete) from other layers (or individual layers, if the aggregate flag is on)

I also added a function to clear any fit line from a given viewer. I think this should give use the flexibility to do whatever we want with the fitting. I suspect we won't end up using a decent subset of this, but this gives us options which we can pare down later.

heywooddogwood:
Checked it out, looks good to me.

Issue #25, opened 2021-08-11
Additional requests for simulated Hubble data displays


patudom:
This relates to the code for generating simulated Hubble data in #4.

- Age calculations in "HubbleSummary_Overall.csv" should be fixed, per our conversation today.
- Please also add an additional output file called "HubbleSummary_All.csv" that concatenates all the outputs of "HubbleSummary_Class_<n>.csv" (without duplicating the headers). (I did this by hand and can submit a PR to add my file to the output folder, but it would be nice to have it generated automatically if we re-run the simulation). 

- Screens 1 & 2 should display 1 student's data
- Screen 3 should allow options to display 1 student's data, their class's data, and all public data from the simulation.
- Screen 4+ should display the histogram data we discussed today (in figure below).

<img width="2507" alt="2021-08-11 Histogram schematics" src="https://user-images.githubusercontent.com/12750048/129112818-ceb2db95-10f2-4e85-b4c6-b2255caca97e.png">

PR #26, opened 2021-08-11
Hubble simulation review

patudom:
This code has plotly versions of the histogram views we will want to include (as glue bqplots) in the app. This is just for reference and doesn't impact anything in the main app. 

Also, see #25 for more notes.


Carifio24:
I'll add in the automatic creation of the class data files to the simulation script in an upcoming PR.

PR #30, opened 2021-08-12
Housekeeping

patudom:
This should not cause any noticeable changes. Just tidying up issues on my github fork from before I learned to put new commits on their own branches.


nmearl:
You may want to rebase to get rid of the "Merge ..." commits. (It doesn't do anything other than keep the git history clean, but something to consider in the future.)

PR #39, opened 2021-08-27
Fix issue with line marks not panning/zooming

Carifio24:
While investigating the cause of https://github.com/glue-viz/cosmicds/issues/36, I noticed that the line marks currently don't pan/zoom with the data in the glue viewer. This PR fixes that issue.

For the scatter viewers, this was a simple as using the same `Scale` objects as the viewer does. However, for the histogram viewers, using the same scales had a small, but visually noticeable, effect on the y-axis scale (particularly in our default positioning, with the minimum of the y-axis at 0).

To work around this, I have the line marks on the histogram viewer use new `Scale` objects built on the viewer's scales, and attach them to the viewer's `bqplot.PanZoom` interaction when appropriate, so that the marks will pan/zoom with the scales correctly.

### DIFF
--- a/cosmicds/app.py
+++ b/cosmicds/app.py
@@ -509,7 +509,7 @@ class Application(VuetifyTemplate):
         # Observe interaction changes so that we can modify when necessary
         # We want to remove any observers that we have previously set
         figure.interaction.unobserve_all(name='next')
-        figure.interaction.observe(lambda changed: self.panzoom_interaction_update(changed, viewer_id), names=['next'])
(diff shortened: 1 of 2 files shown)


patudom:
@Carifio24 This is great - thanks for catching this! One thing to be aware of—I noticed that the fit line ends at ~1000 Mpc. Once we include professional science data, we will likely have data that extends beyond that distance range, although at a redshift of ~0.5 (d~1,300 Mpc), that's where you start to see evidence of dark energy in SN 1a data, so we will likely need to account for that and display the data/models differently at that point anyway.

PR #40, opened 2021-09-01
Recreation of cosmology models from Riess et al 1998

patudom:
Looking ahead to SN Ia science data to think about what visualizations we may eventually want students to work with, and what we need to be doing now in the HubbleDS app to prep students to understand those.


heywooddogwood:
Looks good

PR #41, opened 2021-09-03
Histogram scatter sync

Carifio24:
This PR adds synchronization between the selection in the first histogram viewer on page 4 (that shows data for the students' class), and a subset in the scatter viewer on the same screen. Since the selection in the histogram is along the age component of the summary data, which we can't link to the individual student measurements, this is achieved via a custom listener (`HistogramListener`) descended from `glue.core.HubListener`.

Additionally, since a student won't have access to the standard glue UI, they won't be able to change the active subset in the usual way. To get around this, I've augmented the` activate` and `deactivate` methods for the x-range selector on the histogram viewer to set the active subset to the (only) subset defined on the histogram viewer when the tool is activate and unset the active subset when the tool is deactivated. The function that I used to do this is `extend_tool`, which was added to `utils.py` - I anticipate that we may need to do this for some of the other viewers in the story.

**Edit:** I also added a line fit to the student's selected subset (that vanishes when the selection is empty/deleted), as well as a button to let them clear their selection (necessary since they don't have access to the glue UI that one would typically use to do this).

### DIFF
--- a/cosmicds/app.py
+++ b/cosmicds/app.py
@@ -583,9 +583,7 @@ class Application(VuetifyTemplate):
 
     def vue_clear_histogram_selection(self, _args):
         toolbar = self._viewer_handlers['class_distr_viewer'].toolbar
-        if toolbar._default_mouse_mode:
-            toolbar._default_mouse_mode.deactivate()
-        toolbar.tools['bqplot:xrange'].deactivate()
+        toolbar.active_tool = None
         self._histogram_listener.clear_subset()
 
     # These three properties provide convenient access to the slopes of the the fit lines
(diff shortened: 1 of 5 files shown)


heywooddogwood:
The described functionality seems to work for me.

Some related things I noticed:
- If you first select within the scatterplot using Glue features (which I can't seem to unselect), this undermines what happens when you select in the histogram after. Shall we turn of Glue selector for scatterplot?
- Didn't My value and Class value used to display a vertical line with age estimates in the histogram? It's not activating anything for me.
- Will the data available in the scatterplot change if we switch from MY CLASS to ALL DATA? Do we want it to?
- Selection tool for ALL DATA histogram doing some unexpected stuff that I don't understand.
- Interesting that the color of the selection changes after every time you use CLEAR SELECTION. It's fun, but perhaps confusing to a student? just a thought.

PR #47, opened 2021-09-23
Add functionality for students to 'draw' line

Carifio24:
This PR adds functionality for the student to 'draw' a best-fit line on screen 3 using their mouse. This works as follows:

- When the student first clicks the "Draw Fit Line" button, a listener is added to the scatter plot. This will follow the student's mouse inside the plot and create a line connecting their mouse location to the origin as they move.
- When the student clicks, this functionality is turned off, and their line stays in place. This click also deselects the "Draw Fit Line" button.
- When the student next clicks "Draw Fit Line", a circular point is added to the end of their line. If the student drags this point, they can adjust their line again. Until they do, their previous line is unaffected.

I think this is pretty close to what we had discussed the other day, but I can make changes to this if we think that this user experience isn't straightforward enough.

### DIFF
--- a/cosmicds/app.py
+++ b/cosmicds/app.py
@@ -2,7 +2,6 @@ from os.path import join
 from pathlib import Path
 
 from astropy.modeling import models, fitting
-from bqplot_image_gl import LinesGL
 from echo import CallbackProperty
 from echo.core import add_callback
 from glue.core.state_objects import State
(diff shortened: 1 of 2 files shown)


patudom:
I'm merging this in, so we can have the basic functionality as we continue to develop the student experience, but we still need to adjust the ux so the user doesn't have to click back on "draw a fit line" button to grab & move the line end point.

PR #52, opened 2021-10-07
Add functionality to data tables

Carifio24:
This PR takes the tables from #50 and integrates them with a modified version of the interactive table component from #45. On screen 3, the table consists only of 'completed' measurements where both a distance and velocity value are present. On screen 3, the table is connected to the scatter viewer - making a subset in one of them will be reflected in the other.

 In an effort to simulate how we'll eventually get data from the student (one data point at a time), the data is now populated using the "Select Galaxy" and "Estimate" buttons. This data is currently just in the `app.py` file. Obviously this isn't how the final product will be, but I tried to separate out the "get a new data point" and "do the necessary glue stuff with it" functionality, so the latter will hopefully need minimal changes later.

### DIFF
--- a/cosmicds/line_draw_handler.py
+++ b/cosmicds/line_draw_handler.py
@@ -159,7 +159,7 @@ class LineDrawHandler(object):
         t3 = y_min / y
         t4 = y_max / y
         ts = [t for t in [t1,t2,t3,t4] if t > 0 and t < 1]
-        t = min(ts or [0]) * 0.99
+        t = min(ts or [0]) * 0.98
 
         return x * t, y * t
 


patudom:
@Carifio24 This is looking great - thank you!  A couple minor suggestions:

- When students "Plot my points," we should include the origin in the range of plotted points in the Hubble viewer.
- I think it would be clearer to replace the button text "Erase fit line" with "Erase drawn line" on Screen 3. (Since the button beneath says "Find best fit," I think we want to distinguish between the drawn fit and the calculated best fit.)

Carifio24:
Both of these issues should be fixed now.

- Now, if the student drags the endpoint out of bounds, it will be moved to the place where their line segment crosses the boundaries of the viewer, regardless of direction (the piece of code that does this determines which boundary was crossed).

- I moved the code that was causing the lag into the `Application` constructor. Glue will complain if one tries to add a size-0 `Data` object to a viewer, so I had been deferring the creation of the `Data` that goes into the screen 3 viewer until the student had one point complete. Now I'm just initializing that object with a single dummy data point when the app is created, and overwriting that point once the student has one of their own. This actually makes the code simpler, as the logic is then the same for each point update.

PR #58, opened 2021-11-01
Minor tweaks and fixups

Carifio24:
This PR makes a few minor tweaks and fixes a couple of things that had broken:
* Adds the student data to the professional data viewer
* On a few of the viewers, the axis labels were simply 'distance' and 'velocity'. This PR changes them to 'Distance (Mpc)' and 'Velocity (km/s)'
* In the first comparison screen, change "All data" -> "All Participant data"
* The pan/zoom functionality for the line marks seems to have gotten removed at some point, so this PR re-adds that
* * The histogram/scatter sync broke at some point when the layout changed, so I took this opportunity to fix that and streamline the helper classes that make it work


Carifio24:
@patudom These newest commits fix the slow updating when data points are added to the table - on my machine, there's no delay. I also updated the histogram axis labels, in addition to the scatter ones.

There's some issue with the `tox` tests that I haven't quite worked out yet. I think the root of the problem is that tox is getting `pywwt==0.9.0`, rather than `pywwt==0.13.0` (which is what was used on the last commit that passed). Based on the error that's popping up when I try forcing `pywwt > 0.12.0`, I think it's related to https://github.com/pypa/setuptools/issues/2849, but when I tried restricting `setuptools` to be less than 58.5, I still got the same error.

Carifio24:
With https://github.com/WorldWideTelescope/pywwt/pull/317, our `tox` issues are resolved.

PR #60, opened 2021-11-18
Add MathJax to the CosmicDS app

Carifio24:
This PR adds MathJax, along with a hack to add in inputs from [here](https://groups.google.com/g/mathjax-users/c/Udj9oD9uyU8/m/P3D3NkEGAwAJ?pli=1), to the app.

As for how this is done, in the app's `mounted` hook, we first set up the MathJax configuration for inputs, then fetch the necessary MathJax code from the [jsDelivr](https://www.jsdelivr.com/) CDN.

The inputs that we put into the MathJax formulae can get an identifier (which becomes their `id` in the DOM), classes (if you want multiple classes, just space-separate them), and initial content (if we want it). In a formula, this would look like `\input[my-id][my-class]{Default text}`. The signal for MathJax that something should be rendered as a formula is to enclose it in one of the LaTeX delimiters, with the exception of single dollar signs - see [this](http://docs.mathjax.org/en/latest/input/tex/delimiters.html).

I styled the inputs to have a 1px black border with rounded edges, since I was having a hard time seeing them on the page without it. I would imagine that the Vuetify classes should work fine. We can also use our own CSS for the inputs, provided that we mark it as `!important`.

If you want to see an example, stick the element below into `app.vue` somewhere:

```
<p>
  $$ x = \frac{\input[my-id][my-class my-class-2]{} - 1}{\input[my-id-2][my-class]{}} $$
</p>
```


Carifio24:
@patudom I added in a piece of code that should fix the issue we were discussing earlier.

What it does is set an observer on the app's root element that's triggered whenever the DOM structure changes (i.e. if an element is added/removed). This observer looks at the HTML for each element and determines whether or not MathJax needs to typeset it (basically, it checks if it has one of the LaTeX delimiters), and typesets whatever elements it decides needs it. The alternative would be to specify a list of IDs of elements that we know will need typesetting, and check the elements for those IDs. The way it's done here is likely less efficient (it needs to check the HTML of every added item, rather than just grab the ID), but requires no maintenance as we add/remove elements.

patudom:
Thanks @Carifio24! Looks good, and I don't notice any lag, so I suspect this fix will be fine for our purposes. I agree it will help not to have to remember to add extra flags whenever we put an equation into a buried element.

patudom:
Just to keep all the info in one place:
- To resize the input box, edit the content specified in app.vue, under "input {}".
- The content in square brackets (\input[][]{}) is as follows, from Jon: if you do, e.g. \input[answer-box-1][some-class]{}, you end up getting \<input id="answer-box-1" class="some-class"\>

PR #62, opened 2021-12-15
Line labels

Carifio24:
This PR adds legend labels for both the fit lines in the scatter plots, as well as the vertical lines in the histograms. The labels match the color of the lines in the viewers, and the order of the labels matches the relevant selection order.


patudom:
I'm not sure where the upside down check marks in front of the numbers are coming from, but I'm guessing @heywooddogwood may want to style those legends differently. Merging in for now, so we can still have the functionality and we can work on the design details later.
Screen Shot 2022-01-06 at 5 24 25 PM

PR #65, opened 2022-01-12
Revert "Add measuring tool"

patudom:
Reverts glue-viz/cosmicds#61

I didn't mean to do this merge, and I don't think it went correctly because the sample galaxy list in the table in the merged main repo is incorrect. Hoping this revert and a new merge will fix.


patudom:
(Sounds like the issues is that I accidentally merged PR61 first, then PR64, which overwrote the data updates we needed from PR61. We are going to revert everything and start over with both those PRs.)

PR #70, opened 2022-01-18
Improvements to measuring tool

Carifio24:
This PR makes a few tweaks and improvements to the measuring tool. These changes are:

* If the view in the tool's WWT viewer is changing (moving or zooming), measuring is disallowed. The WWT widget doesn't have a built-in notion of 'view is changing', so I've implemented this by monitoring the RA, Dec, and FOV of the widget. When the widget gets an update, if any of these values has changed, the view is considered to be changing.
    - We aren't always guaranteed to get an update from the widget if the view doesn't change (in particular, this happens when the student clicks on a galaxy and the movement is via the widget's `center_on_coordinates` method). To get around this, the measuring tool keeps track of when it last got an update from the widget. If that time is longer than some threshold time (currently one second), the tool interprets that as the widget view not changing.
* The measuring tool will jump directly to a galaxy starting with the third galaxy the student selects (currently, this is set to the fourth galaxy).
* If the student's measured angular size exceeds half a degree, they get a warning, and the value is not added to their data set. On the backend, `state.warn_size` is set to True. Currently a `<p>` element with red text displays this color.
* Modified the titles of the tables on the second screen (where the tool is) as well as the third screen.

Finally, while v0.15 of pywwt is released, there seem to be some issues with using it in the app. Until I figure those out, I've set `pywwt <= 0.14` in `setup.cfg`.

### DIFF
--- a/cosmicds/app.py
+++ b/cosmicds/app.py
@@ -87,7 +87,6 @@ class ApplicationState(State):
     vels_total = CallbackProperty(0)
 
     haro_on = CallbackProperty("d-none")
-    galaxy_dist = CallbackProperty("")
     galaxy_vel = CallbackProperty("")
 
     calc_visible = CallbackProperty("d-none")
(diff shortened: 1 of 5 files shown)


Carifio24:
@patudom The two most recent commits should fix the issue that you noticed yesterday, with the measured distance failing to display in the input box.

PR #71, opened 2022-01-24
A Collection of UX/UI Updates

heywooddogwood:
Updates include:

- Moving "select a galaxy" and having table show up a step earlier.
- Turn "select a galaxy" bursts into a single burst for which the language changes as you click
- Create consistent look for placeholder buttons that will be removed as functionality gets implemented
- Add x-out buttons to all of the pup-up dialog windows
- Pull information and reflection pop-up dialogs out of the bursts so they remain accessible
- Add back button to all bursts

### DIFF
--- a/cosmicds/app.py
+++ b/cosmicds/app.py
@@ -97,6 +97,8 @@ class ApplicationState(State):
     
     spectrum_tool_visible = CallbackProperty(0)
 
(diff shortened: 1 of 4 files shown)


heywooddogwood:
Pat's changes check out on my computer.

PR #72, opened 2022-01-24
Velocity calculation

patudom:
@Carifio24, here is my start to the dialog with the velocity calculation input boxes (under data/vue_components/reflect_velocity_windows.vue)

The input boxes that need to be stored and validated on the back end are \input[lam_obs] and \input[lam_rest]. If you can please write up an example of how to work with those values, that would be helpful.

The desired behavior:

- If input values don't match values in table, student is prompted to try again. (and if check fails a 2nd time, we probably want to prompt them to check with a neighbor or their instructor).
- If input values match, they are displayed on the next dialog screen (in place of text that says "this should display student's value of lam_<obs/rest>")

### DIFF
--- a/cosmicds/app.vue
+++ b/cosmicds/app.vue
@@ -178,6 +178,17 @@
                               cols="6"
                               class="galtable_column"
                             >
+                            <!-- Temporarily moving up to work on equations-->
+                                    <reflect-velocity-windows
+                                      button-text="reflect"
+                                      close-text="submit"
+                                      @close="
+                                        console.log('Submit button was clicked.');
+                                        state.rv1_visible = 0;
+                                        state.calc_visible = 'd-block';
+                                      "
+                                    >
+                                    </reflect-velocity-windows>
                               <!-- GUIDANCE ALERT - introduce students to WWT Viewer -->
                               <v-alert
                                 :class="state.marker == 'exp_sky1' ? 'd-block' : 'd-none'"
--- a/cosmicds/app.vue
+++ b/cosmicds/app.vue
@@ -313,6 +313,25 @@
                                       class="black--text"
                                       elevation="2"
                                       @click="
+                                        state.marker = 'exp_sky1';
+                                        state.galaxy_table_visible = 1;
+                                        state.gal_snackbar = 0;
+                                        state.dist_snackbar = 0;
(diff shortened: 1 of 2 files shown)


Carifio24:
After thinking about this some more, I think that since we aren't concerned with storing students' incorrect responses, that we can do any validation on the JS side. As a simple example, let's say that we wanted to verify that the value within some input is within 3% of the expected value. We can add a method to our Vue component to do this calculation. In the code, this makes our reflect windows component look like:

```
<script>
module.exports = {
  props: ["buttonText", "titleText", "closeText"],
  data: function () {
    return {
      step: 1,
      dialog: false
    };
  },
  computed: {
    currentTitle () {
      switch (this.step) {
        case 1: return 'Observed vs. Rest Wavelengths'
        case 2: return 'How Galaxies Move'
        case 3: return 'Calculate Galaxy Velocity'
        default: return 'Complete'
      }
    },
  },
  methods: {
    validateWithinPctTol(inputID, expected, pctTol) {
      const input = document.getElementById(inputID);
      const value = parseFloat(input.value);
      return Math.abs(value - expected) / expected < pctTol / 100;
    }
  }
};
</script>
```
Basically, we grab the input using its ID, parse its value, and compare to what we expect. You can then call that inside an event handler in the HTML. If we wanted to do this check when the next button is pressed, we can do this by changing the next button's `@click`. This would look like
```
<v-btn
  :disabled="step === 5"
  color="accent"
  text
  @click="() => {
    const valid = this.validateWithinPctTol('lam_rest', 5177, 3);
    if (valid) {
      step++;
    } else {
      // do whatever - i.e. make some warning message visible
    }
  }"
>
```

If we find that we're reusing a lot of types of validation (which seems likely), then we can refactor and put these in some utility file. The only other issue is how to get the expected values into the dialogs. I have a few ideas on this, but I'd like to see what the state setup looks like in the refactored app first.

patudom:
This branch is old and stale. Migrated relevant content by hand to new branch (velocity-dialog-v2) and closing this one.

patudom:
I didn't actually see the content from this branch in velocity-dialog-v2, so I deleted that branch and am adding this manually to the new main on branch velocity-calc-new-main.

PR #74, opened 2022-02-01
WWT galaxy selection and screen 1 connections

Carifio24:
This PR adds the ability to select galaxies using the WWT widget, and adds some necessary connections between elements on Screen 1.

The list of changes includes:
* The data used for the student measurements has been changed to the SDSS data
* The spectrum data has been added to the repository. This is a decent amount of data (~240 MB), and significantly increases the size of the repository. I think it's easiest to keep it in here for now, but we can think about how better to handle this in the future
* As we talked about last week, I removed the exploration phase of the WWT viewer from screen 1 so that it can go in the intro. This PR contains code for a `GalaxyExplorationTool` that we can place there
* The students can now select galaxies using the WWT widget on screen 1. The student's selected galaxies show up as a pinkish color to distinguish them from unselected galaxies. These selections are linked to the table on screen 1
* The table component has been modified to listen to not show checkboxes, and to instead listen to `click:row` events. This frees up some horizontal space on the left-hand side of the table
* When a student selects a galaxy, the relevant SDSS spectrum data is loaded into the app. When the student selects a row in the galaxy table, that data is loaded into the spectrum viewer
* When the student clicks while mousing over the spectrum viewer, their measured wavelength is added to the galaxy table
* For the purposes of testing, I've added a "Find Velocity" button that becomes active when the student has measured a wavelength for the currently selected galaxy. It adds a velocity and rest wavelength value in the appropriate table row
* Currently tests are failing due to some issue with testbook and how the pywwt Jupyter widget detects the running Jupyter kernel. I'm still working on hashing this out, but I don't think we need to hold things up for that


patudom:
pyWWT 0.15 seems a little bit temperamental on Pat's computer. It works on and off, and @Carifio24 and I can't figure out why. Jon will try to track this down, but we are merging it for now because it works most of the time.

PR #79, opened 2022-03-04
Updates to intro wwt content

patudom:
- format WWT widget & buttons provided by @Carifio24 & add to slides 3-5.
- Minor edits to text
- Some reformatting per suggestions from Alyssa.

Still needs images, they weren't working, so will wait to see @heywooddogwood's Stage 1 dialog to see how that's done.

### DIFF
--- a/cosmicds/app.py
+++ b/cosmicds/app.py
@@ -30,7 +30,8 @@ from .components.viewer_layout import ViewerLayout
 from .components.widget_layout import WidgetLayout
 from .histogram_listener import HistogramListener
 from .line_draw_handler import LineDrawHandler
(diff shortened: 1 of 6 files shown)


Carifio24 (approved):
Aside from the missing image, everything looks good to me! The new functionality all seems to be working smoothly.

@patudom The problem with the image is that the URLs don't need to be wrapped in `url()`. That's a CSS function, but you don't need it for the `v-img` HTML attributes. If you take those out, the image shows up, but it's pretty small so I'll let you resize it.

patudom:
Thanks, Jon. I originally had it without the url() wrapper, but I think there were other issues that cause the image not to appear. All fixed now.

PR #80, opened 2022-03-09
Ux updates 4 - Spectrum Slideshow

heywooddogwood:
This Pull Request is somewhat incomplete, but provides an update to the slideshow dialog explaining and exploring emission/absorption/spectra. More edits needed to update the appearance of the images and the layout of the slides to appropriately adjust to various screensizes. Also, a couple slides have yet to be built in.


patudom:
Thanks, @heywooddogwood. This is looking good, and I'm glad to have the examples of how you added the images and laid out the slides, so I can copy the methods in the intro slideshow. I don't think this dialog will cause conflicts anywhere else, so I will leave this unmerged until you finish the updates. 

A few notes:

- I think the blue box on the 1st slide probably wants to be above the rest of the slide text because I'm not sure it makes sense read in the reverse order.
- If possible, I think we should try to keep the blue box text in an identical location/layout on slides 1-3, so it's clear that it's the same text on all 3 slides, so the student doesn't try to keep re-reading it over and over again, but it's there for reference if they need it.
- On Slide 1, I'm not seeing an actual em dash. Rather, I am seeing "&emdash;". Not sure if it's just me.

patudom:
@nmearl We noticed that Voila is having an issue where it doesn't know how to find the images within our directory structure. Any thoughts on how to point Voila to the correct locations? Thanks!

PR #81, opened 2022-03-15
Update grid style

heywooddogwood:
We've got a new layout here, in line with Alyssa's suggestion from the first couple slides of the mockup google slides file (at least, they are the first couple slides currently). The new layout puts the bursts in a right hand column that move up and down with the relevant widget to focus on at any given point. Functionality regarding flow from widget to widget and step to step is maintained as before. Strictly cosmetic.

Although I did accidentally add an additional burst to split up the Select 5 Galaxies burst. I'll include that in the main as well.

# Pull Request Template

## Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

### DIFF
--- a/cosmicds/app.vue
+++ b/cosmicds/app.vue
@@ -174,14 +174,8 @@
                           class="pt-0"
                         >
                           <v-row>
-                            <v-btn
-                              color="green"
-                              @click="select_galaxies()"
-                            >select 5 galaxies</v-btn>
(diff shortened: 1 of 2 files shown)


patudom:
Looking good. There will be more layout updates to come (swapping text boxes from right to left), but merging for now because we like this better than what we had before.

PR #83, opened 2022-03-16
Ux updates 7

heywooddogwood:
A series of UX updates that includes the following:

- Put the instructional bursts on the left column, the widgets on the right.
- Switching the intro slideshow GET STARTED button so that it is for internal use and will be deleted except on the final slide.
- Delete the helper alerts at the bottom of the page
- Restyle the Messier Buttons, as well as the Pan/Zoom instructions.


patudom:
There's some inconsistent behavior of the Messier object buttons on slide 4 vs 5 on my computer
Screen Shot 2022-03-16 at 1 06 59 PM
Screen Shot 2022-03-16 at 1 07 05 PM

PR #84, opened 2022-03-18
Initial migration to new architecture

Carifio24:
This PR does the initial migration of the content from the current main branch into the refactored architecture. At the moment this mostly consists of stage one, with some of the stage two content, but I'll be adding more soon. Overall this was a relatively straightforward process, but I'll talk about some of the key things I did and things that I think will be helpful moving forward.

As @nmearl and I discussed, I aimed to keep as much functionality as possible contained within components. As things are now, the stage-level Vue files are largely for dictating the overall layout and when certain components should or should not appear. Pretty much every block of content now lives within its own component file. I've also tried to move everything that I think is specific to the Hubble data story inside the `cosmicds/stories/hubbles_law` folder, but I'm sure there's more room to improve what goes where.

A lot of these components, whose content lived in `app.vue` before, only interact with the state in a pretty trivial way that can be done entirely in JavaScript - either using a state property to determine the layout or setting a state variable after some action. Since these components now need some way to get access to the state, I've created `GenericStateComponent`, which can serve as the Python backend for Vue components that only need frontend access to some state variable. See the example usage of this in stage one - you just need to give it the state you want to use, and the filename and location of the template file.

In order to make the glue viewers look like they belong in our application, I've created a custom glue toolbar using the styling that we've been using for our viewers. This should work much better than the current approach, which hides the glue toolbar and hotwires its buttons to the viewer tools. You can see this toolbar in `cosmicds/components/toolbar`. Currently the toolbar uses material design icons as the button images, which are set via a dictionary in its file. The spectrum viewer now uses this as its toolbar, and in `cosmicds/viewers/cds_viewers.py` I created 'CosmicDS-ified' versions of the scatter and histogram viewers, which is done via a class factory that we can easily use for any other glue viewers.

The rest wavelength functionality in the spectrum viewer has been moved into its own tool. I've also created an "info tool" for viewers - it doesn't do anything by itself, but it exposes a `active` traitlet that we could listen to for showing/hiding informational dialogs.

In the current master, the basic control flow, at least in screen 1, is dictated by the state's `marker` variable. I've largely retained that here, but have hooked up these marker's to the stage's steps. Basically, markers are used here as intermediate waypoints between steps. When the marker reaches one of the values defined in the state's `step_markers` member value, the stage's step index moves forward, which shows in the left-hand side menu. Similarly, if the student clicks on a step in the side menu, the state's marker is set appropriately. Currently, a student cannot select a step if they have not completed the preceding step. For ease of testing, I've left every stage selectable for the moment.

I hope this all makes sense. Let me know if there are any questions/comments. I've tried to keep this synced up with the current master as best as I could, but it's possible that I missed some stuff along the way.


patudom:
@Carifio24, WWT is still giving me a 404 error in Voila. Any tips on troubleshooting this?

Carifio24:
Looking at `setup.cfg`, it looks like I forgot to add `wwt_kernel_data_relay` to the install requirements. Does installing that help at all?

patudom:
Doesn't look like it. These are the error messages I'm getting in case it's useful:

    [Voila] ERROR | unhandled iopub msg: wwtkdr_claim_key
    [Voila] ERROR | unhandled iopub msg: wwtkdr_claim_key
    /Users/Pat/opt/anaconda3/lib/python3.8/json/encoder.py:257: UserWarning: date_default is deprecated since jupyter_client 7.0.0. Use jupyter_client.jsonutil.json_default.  
     return _iterencode(o, 0)  
    [IPKernelApp] ERROR | No such comm target registered: jupyter.widget.control  
    [IPKernelApp] WARNING | No such comm: 8beee762-7851-4961-ac86-bbc7319af534`

patudom:
@Carifio24 helped resolve this issue. When I compiled the new version of the app, it appeared to grab @nmearl's updated version of voila from https://github.com/nmearl/voila, but when we checked what was actually on my computer, my voila didn't have any of Nick's fixes to make pyWWT work.

Solution, in case anyone else runs into this problem:

- pip uninstall voila
- recompile the cosmicds app

nmearl:
Thanks for troubleshooting the installation issue, @patudom!

patudom:
Also, it appears that the run tests are failing, but @Carifio24 doesn't think the failed tests are actually impacting functionality of the app. We know it's not a great practice to allow the failed test, but in the interest of making forward progress, I'm going to merge this PR and we will sort out the failed test later.

PR #85, opened 2022-03-18
Update bqplot styling

patudom:
This is just a start to the bqplot style updates. We will continue in the refactored app. To be fixed still:

- For now, this makes all the plot labels white, so they will disappear in light mode. We need to figure out how to trigger style updates for all the viewers when the darkmode is toggled.
- The spectrum viewer tickmarks seem to be white only for the first spectrum you plot. Once you switch to other galaxies with different y-axis values, the new values are displayed as default gray.


patudom:
I'm confused about why this is bringing in commits from 8 months ago. @Carifio24, could this be related to our organization migration?

Carifio24:
@patudom Just curious, how long ago did you make this branch?

patudom:
@Carifio24 As far as I know, ~2 days ago, unless I reused a branch name from way back when that I'd forgotten about?

I did do a merge to this branch from upstream main, but I do that pretty often and have never seen this before.

patudom:
This was on old-main. Moved functionality over to new main in #92

PR #88, opened 2022-03-22
Put back @heywooddogwood's updated spectrum info dialog

patudom:
Plus:
- fix em dash on slide 1 which was appearing as "&emdash;" on my computer. (@heywooddogwood, hopefully that didn't break the emdashes for you.)
- move blue boxes of text on slides 1-4 to top
- even up column width of text on slides 1-4 so blue box stays in same location.

Issues:
- Voila doesn't know how to find the images. They only appear when app is run in jupyter notebook.
- Vue doesn't seem to be observing "no-transition" and is swiping the text in from the right for slide changes.


patudom:
Images are now displaying properly in both Voila and Jupyter Notebook after merging in #89 to this branch.

heywooddogwood:
Everything looks good on my review! Some unsuccessful checks reported, but it sounds like that's not an issue on our end, but rather github.

PR #90, opened 2022-03-25
Ux rebuild 1

heywooddogwood:
Updates included:
- Remove page-number avatars from dialogs
- Clean up browser width layout thresholds
- Move calculate velocity next to INFO button
- Figure out why stage 1 scaffolds drop out.
- Recreate 2 column layout
- Add Light/Dark theme toggle


Carifio24:
@heywooddogwood @patudom To better fit with the changes made in #89, the dark mode toggle now observes the `dark_mode` property of the `ApplicationState`, rather than a property on the `Application` itself. I was mistaken the other day - the stages actually do not have a reference to the application object. However, they do now have a reference to the application state, so we should use that. Just wanted to let you know why I made a change.

PR #92, opened 2022-03-25
style spectrum viewer - works only in dark mode so far

patudom:
Bring over functionality from old-main PR85 to new main and will close #85.

Keeping in draft mode for now while we iron out these issues:
- For now, this makes all the plot labels white, so they will disappear in light mode. We need to figure out how to trigger style updates for all the viewers when the darkmode is toggled.
- The y-axis tick marker numbers are wonky. They are fine on the first galaxy, but when you switch to other galaxies with different y-axis values, only the numbers that were on the original galaxy plot appear.


patudom:
This is being superseded by Jon's version.

PR #94, opened 2022-03-28
Stage 2 slides

LilyN11:


patudom:
Remaining issues:
- I can't figure out how to make the formatting work in the mc-radiogroup feedback options in slide 7. The 2nd question has <strong> tags that just get treated as a string. @Carifio24, do you know if there is an easy fix?
- We are waiting for Anna to provide a better version of the image on slide 6.
- The image layouts need work.  @heywooddogwood, please fix things like images in the same row that have different heights and/or redo the layouts however you see fit.
- The equation layouts are not robust with respect to different screen sizes. We forced the orange equation boxes not to run across the whole screen by specifying a column width, but the equations will spill out of the boxes if a user makes the screen sufficiently narrow. I don't know how to make the card lengths more flexible.

patudom:
Hmm. Github noted some merge conflicts, so I fixed them on the GitHub site for the most part by just accepting the versions coming from “main” instead of PR 94. When I pulled down the merge commit and ran the app on my computer, the spectrum graph is back to the scatter plot instead of the line graph. I think this means I did something wrong in the merge commit? @Carifio24, would you mind taking a look when you get a chance, please? Thanks!

heywooddogwood:
Everything seems to be functioning fine. There are formatting updates to be made, but those will appear in a future branch/PR.

heywooddogwood:
Things to be done:

- turn off sliding slide transitions that still occur for several of the slides.
- math slideshow container formatting to that from Stage 1 (e.g. header color)
- Make text in alerts/colored cards white
- Center MathJax equations
- Fix width of cards to have margins
- Get MC to match format from Stage 1
- Additional cleanup as needed

PR #96, opened 2022-03-29
Ux rebuild 2

heywooddogwood:
A collection of changes largely connected to the scaffolds:
- get rid of NEXT button for choose row
- use v-if for conditional visibilities
- all widgets hold borders at all times
- break rest wavelength scaffold in two
- break select galaxies scaffold in two


patudom:
I'm not sure where my commit [eb83854] is coming from, but we think it's ok so going ahead and merging.

PR #98, opened 2022-03-30
Make dialogs draggable

Carifio24:
This PR adds functionality to make dialogs draggable throughout the app. This is accomplished via some JavaScript added into the app's `mounted` hook. This code is originally based on [this](https://github.com/vuetifyjs/vuetify/issues/4058#issuecomment-450636420) comment, but I made a few tweaks:

* You can't drag the dialog outside of the current window, so the only way to end up in that situation is a window resize. Thus, rather than use the `setInterval` call, I moved the out-of-bounds check into a `ResizeObserver`.
* I added some CSS to handle the grab/grabbing cursor on the dialog title. Unlike in the comments below the linked one, the grabbing cursor is related to a `dragging` class that gets un/set in the JS.
* While the dialog is being dragged, `v-overlay` elements are hidden (so background elements won't be grayed out).

As we talked about in the dev meeting, the selectors rely on DOM classes that Vuetify gives to its elements. However, if we pin to a particular version of `ipyvuetify` (and thus Vuetify), this shouldn't be an issue.

I'll explore options for modal vs. non-modal dialogs, but I think this is a good starting point for this functionality.

### DIFF
--- a/cosmicds/app.vue
+++ b/cosmicds/app.vue
@@ -308,7 +308,7 @@ export default {
         d.el.style.position = "fixed";
         d.el.style.margin = 0;
         d.oldTransition = d.el.style.transition;
-        d.el.style.transition = "none"
+        d.el.style.transition = "none";
         d.title.classList.add("dragging");
         d.overlays = document.querySelectorAll(".v-overlay.v-overlay--active");
         d.overlays.forEach(overlay => overlay.style.display = "none");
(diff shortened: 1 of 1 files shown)


patudom:
Thanks, Jon! This looks good and I think may be good enough for the pilot. If we aren't able to get the nonmodal dialog working, hopefully students can figure out that they can close the dialog, scroll the main page in the background, and reopen the dialog. 

@heywooddogwood, this will really only matter for the calculation dialogs, but we'll need to make sure they are small enough that you can move them out of the way to see the info you need underneath. The spectrum info dialog covers up most of the background, wherever I drag it.

PR #100, opened 2022-04-07
Ux updates 10

heywooddogwood:
A short history of commits in this PR:

- Add two missing slides to the spectrum slideshow
- Revamp theme colors and keyword assignments
- Update header gradient
- Restyle the slideshow/viewer/widget headers
- adjust font sizes


patudom:
From an accessibility point of view, I think the wider width of the highlighted widget borders is probably better. It's much less noticeable now, and even though the thinner border is more attractive, I think it's more important for the highlight to be easy to see. Other than that all changes look good.

PR #101, opened 2022-04-08
Migration part 2

Carifio24:
This PR does some more of the migration from the old architecture to the new one. This primarily consists of adding some of the viewers from stage 3, but there are also some improvements to the table component and the spectrum viewer. These connections don't all work quite right yet, but I wanted to start getting some of this content into the main branch.


Carifio24:
@patudom Per our conversation earlier, I re-enabled the user selection line on the spectrum viewer when the rest wavelength tool is active. I also temporarily disabled the `HistogramListener` - that seems to be related to the loading issue we were seeing before. I'll sort it out and re-enable it in the next PR.

Carifio24:
@patudom @heywooddogwood I was able to replicate the issue with the app not loading in voila on my MacBook, and was then able to get it working by changing some of the timeout settings in voila. Can you try this latest commit and let me know whether it works for you?

Carifio24:
@patudom As we discussed, I've commented out all of the functionality of stage three. If this lets this run for @heywooddogwood, then I think we can merge this in now and deal with the timeout issue a bit later.

PR #103, opened 2022-04-16
Updates to scaffold alerts for spectra and start functionality for validating velocity calculation

patudom:
Keeping this as a draft until I make more progress on the velocity calculation dialog.

- Bring over MathJax colors & input box styling from branch on old main that was never merged.
- Implement @Carifio24's MathJax input scheme laid out [here](https://gist.github.com/Carifio24/69fb8a30d72037803b8e0440bc099e85 ). 
- Update Spectrum Dialog button sequencing, picking up from where @heywooddogwood left off in #102.
-- Spectrum Dialog "Learn More" button now does not appear until measurements have been completed. (We may decide to change this, but for now it's hidden until needed, then always accessible after).
-- "Click LEARN MORE button" text in mee_spe1 instruction alert becomes "NEXT" after "LEARN MORE" has been clicked.  (It would be nice to auto advance the alerts when the LEARN MORE dialog is closed, but doing that would put the user back at this same section of the alert sequence every time they open and reclose the spectrum dialog, which would be problematic.)
- Edit Alert text to match mockup slideshow
- Update velocity reflection dialog to match mockup slideshow 

Issues that remain:
- Velocity calculation dialog is hidden right now because it is unfinished. 
- Button in final alert in spectrum sequence still needs to be wired up to wait for all 5 galaxy wavelengths to be measured.
- Down the line, we could make student actions (like clicking to measure wavelength) auto advance some of this sequence, but I don't think it's necessary and is kind of fussy to code.
- Still need to improve the radio button functionality for the last two questions in the velocity reflection dialog which don't truly have right answers.


heywooddogwood (approved):
- Loops back on self within guidance once I add velocity to table
- gave me duplicated rows of the final velocity calculation and the equation (see screenshot)
- not updating table. Perhaps that’s because I auto-selected 5 galaxies instead of choosing them myself?
- process of updating to show previous steps in calculation slightly awkward, though I really like the execution.
- Why do elements show up as H-? until you click on the row?
- I love how the process starts in the scaffolds and then pops out to the dialog once you have the values you need.
- The first jump in the doppler calculation might be the most jarring. Perhaps because it shows so much information in the first step, and then reduces to v=c(## - 1).
- Maybe we put the v, c, lambdaobs etc into a card, and the active calculation just gets an outlined border?
- Love how the multiple choice works in. Maybe we put the answer as black text in an alert frame. Something tells me the green on black will not be 508-compliant
- I like how putting in the speed of light wrong calls out where to look for the right information
- I like how the next button turns to “Try AGAIN”.
- Perhaps next / back / calculate should all be buttons, like TRY AGAIN is.
- need to disable the multiple choice question after they get the right answer. I think? Although maybe it’s not a big deal if we aren’t tracking this datapoint.

PR #106, opened 2022-05-03
Student data generation

patudom:
@Carifio24, this is the start of the offshoot branch that team members can use to quickly generate "seed" student data.

Still needed
- ~~Addition of a field for team members to tag their measurements with their name, in case we need to sort out later who submitted which data points.~~
- ~~Wire up the Stage 2 distance measuring tool as follows:~~
  --  ~~Add "angular size" column to table.~~
  --  ~~When student makes measurement, the angular size and distance should autofill into the table (and update when the measurement is adjusted.)  No additional button click should be needed to calculate the distance (as real students will do).~~
- ~~A "start over" button on the Stage 2 page can be clicked when the measurements are complete. That should clear all the data and create a fresh slate for a new round of measurements with a new dummy student ID tag.~~

Let me know if you have questions. Thanks so much!

### DIFF
--- a/cosmicds/__init__.py
+++ b/cosmicds/__init__.py
@@ -17,11 +17,3 @@ for comp_path in vue_comp_dir.iterdir():
 
 from .stories import *
 from .tools import *
-
-# Monkey-patch pywwt kernel connection function
-from pywwt import jupyter_relay
-
-def _override_compute_notebook_server_base_url():
-    return "/api/kernels/"
-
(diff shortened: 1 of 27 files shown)


Carifio24:
Note to self:
* Add a button that flags weird spectra, so that we can either change the element or remove them.
* If a spectrum is marked bad, replace it with a new one in the data set.
* Calculate age and store in in the student's story state when moving on to next set of data points.

patudom:
Closing in advance of repository restructuring. We will still be able to access and generate new data  while closed.

PR #107, opened 2022-05-03
Migration: Part 3

Carifio24:
This PR contains the next phase of the content migration to the new application architecture. Most of the changes relate to stage three, but there are some changes to stage two as well. I'll go through the main changes here.

* The various scatter and histogram viewers have been added to stage three. There's no flow here yet, but all of the viewers and their connections should be in place.
* The distance measurement tool in stage two is now working.
* All of our dialogs should now be draggable. The ones that weren't had a different title class and so our selectors weren't catching them. To fix this, I've added `v-toolbar__content` as a title bar class to look for.
* The line fitting and line drawing tools are now functional. In this framework I've reworked these to be glue viewer tools which the students activate via the toolbar. The line fitting tool toggles between showing the relevant fit lines and hiding them. The tool is a `HubListener` and will react to changes in the layers.
    - To allow the tooltip of the fitting tool (or any other tool that we create) to change based on its state, the toolbar now checks whether the `tool_tip` field is a `Traitlet`. If it is, it will listen to changes in the tooltip and update it.
* In the table component, the set of selected items can now be either a `SubsetGroup` or just a `Subset`.
* All of the calls to `new_subset_group` in the app setup have been removed - these are now deferred until they're necessary later. This will hopefully make it easier to the app to load on everyone's computer.
* In order to simplify having to hide layers from various created subset groups, and avoid any sort of layers flashing into existence and then being hidden, I've added an `ignore` method to our `cds_viewer` viewer class factory. This method takes one argument, which is a function that should accept a `Data` or `Subset` object and returns a boolean. When the data collection goes to add a layer to this viewer, if any of these conditions returns true, then the layer is ignored. This lets us pre-emptively ignore layers that satisfy certain conditions, rather than needing to worry about layers that get created during the runtime of the app.


Carifio24:
I pushed a few more commits to clean up some minor issues. In particular:
* Fixed a z-ordering issue that was hiding tooltip for the FAB in the distance tool
* Fixed a bug with the line drawing tool's logic for bringing the endpoint back in bounds if a student drags it out
* The line fitting tool will now respond if a new layer is added if it's active. If a layer is changed, the tool now updates (rather than recreating) its line mark, which looks better on the frontend.
* When the student data is updated, the fit viewer on stage three will reset its limits to include all of the data points. I also added a bit of padding in both dimensions (20% of what would otherwise be the max) to the fit viewer bounds

The only issue directly relating to these changes that I haven't worked out yet is making the endpoint of the drawn line draggable when the tool is inactive. This seems a bit tricky because of the way that `bqplot` handles interactions. I'll have to think about that a bit more, but I think the rest of the changes can be merged in without it.

PR #110, opened 2022-05-24
Copyedits 1

heywooddogwood:
The follow sections have been edited for language and student flow:
- Intro slideshow
- Guidelines through Doppler Calculation
- Spectrum Slideshow
- Reflection Slideshow
- Doppler Calculation slideshow

General formatting updates appear with these copyedits as well.


patudom:
Hi @heywooddogwood, this update is looking GREAT. I made some minor edits within the previous 5 commits. Please take a look and let me know if you'd like to discuss any of this further.

I went through the experience in detail and noted the following issues that you and/or I should update. These are in the order I encountered them (not necessarily in order of importance):
- "Notice that your My Galaxies table now has a row for your selected galaxy." 
  -- Need to highlight table row.
  -- Would it make sense to put card next to table, or split it?
- "Your galaxy spectrum has an emission line marked."
  -- After lambda is clicked text should update to "The rest wavelength of your line is now displayed in the graph as a vertical dotted line." rather than remaining the same, which felt confusing.
- Need to add a count down for measuring wavelengths for each galaxy and block student from proceeding until they've all been measured.
- Reflection dialog
  -- When I closed the dialog part way, it did not hold my place
  -- The "Next" button appeared when I closed the dialog without completing it. It should not change to "next" until reflection questions are all complete.
  -- It would be helpful to add a note on the first question, something like: If the dialog box is blocking information you need in the table, you can drag it to a different location using the header bar. You can also maximize your browser window to fit as much as possible on the screen.

heywooddogwood:
Everything looks good. Need to add a few &times; to the doppler equations, but we can get those in on the next round of updates to stage 1.

PR #112, opened 2022-06-01
Functionality improvements

Carifio24:
This PR contains functionality improvements across the app. A summary of the primary changes:
* The tile and spectrum flagging functionality has been added. In this branch, flagging a galaxy here will only remove it from the student's dataset (unlike the data generation branch, where a replacement is automatically populated).
* In stage 1, if the student has already selected a galaxy in the table, the spectrum viewer automatically displays once the student gets to that part of the stage.
* The stage 2 table has been updated. The velocity column has been removed and the angular size column has been added. The galaxy type column now displays the full type name, and that column's title is now "GZ Class".
* The database is now properly updated when the observed wavelength is changed using the spectrum viewer
* A field of view display has been added to the distance tool in stage two. This display shows the FOV of the display, rounded to the nearest appropriate unit (to the nearest degree if the FOV is > 1 deg, to the nearest arcminute if it is < 1 deg and > 1', otherwise to the nearest arcsecond)
* Multiple choices now disable once the correct answer has been selected, and display a score for the question. The default is to start at 10 points and reduce by 2 for each try, but this can be changed on an individual basis
* Updates to the callback system employed by the MC component
* Axis tick marks will now display as the correct color, even when new ticks are created when zooming in. The key was to set the `fill` property of the marks' `css_style`, rather than `color`.


patudom:
> @patudom @heywooddogwood is there a piece of CSS somewhere that's keeping these white?

Yes, in PR109 (https://github.com/cosmicds/cosmicds/pull/109/commits/a7ee7ff8c8aeeb5cdfa8c3938d6bb563431705af), Harry added `color="white"` in cosmicds/components/viewer_layout/viewer_layout.vue to give all the graphs a white background as a quick fix to the label color issue. I'm removing it in my next commit, which will allow your axis label formatting fix to work.

patudom:
The scoring for the MC radio group is still not behaving as expected.  (For example, for the questions on Slide 3 & 4, when I choose 1 incorrect answer, then the correct answer, I expected to get a score of 8 points but keep getting 6. But the question on Slide 7 gives me 8 points as expected). I am going to merge this branch anyway, and we can look into this in a future branch.

PR #113, opened 2022-06-06
Ux updates 14

heywooddogwood:
for pat to check images in slideshows


heywooddogwood:
All of Pat's updates check out.

PR #114, opened 2022-06-07
Stage 1 ux updates

patudom:


patudom:
There is an issue where in select_galaxies_2_guidance, when you click on a galaxy in WWT, state.gal_selected is supposed to switch to True and update the text in the alert to “What do you think of this galaxy?”  Something about the functionality to avoid re-adding galaxies the student already selected broke the trigger for the text change. 

Jon says: "There was an error in the WWT callback when the selection tool's data was empty that was causing the callback to exit before setting gal_selected to True."  Will fix this in a future branch.

PR #115, opened 2022-06-08
Spectrum flagger branch

Carifio24:
This PR contains the spectrum flagger branch, which is specifically for our internal use to check spectra. Pretty much everything except the spectrum viewer has been stripped out. The spectrum viewer will cycle through unchecked galaxies.

As each spectrum appears, one can mark it as "good" or "bad" using the two buttons. These two buttons will send requests that directly update the `spec_is_good` and `spec_is_bad` fields in the galaxy table of the database. When you mark a galaxy, the next one will pop up. If you make a mistake, you can use the arrow buttons to go back and fix it.

This branch is pulling galaxies from a different endpoint than the regular app  (`/hubbles_law/unchecked-galaxies` rather than `hubbles_law/galaxies`). Thus, if you close the app and re-open it, galaxies that have already been checked won't appear in the list.

### DIFF
--- a/cosmicds/app.py
+++ b/cosmicds/app.py
@@ -29,7 +29,7 @@ class Application(VuetifyTemplate, HubListener):
     _metadata = Dict({"mount_id": "content"}).tag(sync=True)
     story_state = GlueState().tag(sync=True)
     template = load_template("app.vue", __file__, traitlet=True).tag(sync=True)
-    drawer = Bool(True).tag(sync=True)
+    drawer = Bool(False).tag(sync=True)
     vue_components = Dict().tag(sync=True, **widget_serialization)
     app_state = GlueState().tag(sync=True)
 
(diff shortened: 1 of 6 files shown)


heywooddogwood:
Doing a round of spectrum flagging now. Will post message when I stop.

heywooddogwood:
Finished for now.

Ending galaxy:
Spec-1321-52764-0148

Everything after Spec-1321-52764-0148 has the same spectrum??

Carifio24:
Hmm, I'll take a look and see what's up with that

Carifio24:
@heywooddogwood So when I start up this branch, spec-1321-52764-0148 is the second galaxy I see in the list (the first is spec-1316-52790-0095). However, I am seeing different spectra for each galaxy - if you load up this branch again, does the same thing happen?

patudom:
Closing this branch now that we are done with spectrum flagging. Leaving the branch undeleted, though, in case we need to revisit any of this content later.

PR #118, opened 2022-06-09
Fix multiple choice issues

Carifio24:
This PR should fix the weird behavior of MC scoring sometimes being wrong and the MC being prematurely disabled. It looks like that was due to a `@mouseup` listener on the MC option label which has been removed.


Carifio24:
@patudom I've implemented the changes that we discussed earlier today. There's now a `scoring` prop on the MC, which is true by default, that will turn off scoring if set to false. Also, if `points` is given as an array and the student exhausts the array length, they will get 0 points.

I also made a small change to correctly show "point" rather than "points" if the score is 1.

PR #119, opened 2022-06-09
Angular size galaxy dos donts slideshow

patudom:
This is the branch where we will create the do's/don'ts slideshow for measuring galaxy angular size.


patudom:
Thanks, @heywooddogwood! @LilyN11, see this branch for the slideshow images and template slides.

patudom:
@LilyN11 We need to merge in this branch so @nmearl can split off the Hubble Data Story into its own repository. This slideshow is working well with other updates that have happened. After Nick finishes the repository migration, hopefully it won't be too complicated to swap out the slideshow images in the new hubbleds repository.

PR #127, opened 2022-07-14
Data and state updates

Carifio24:
This PR takes the first steps towards accomplishing two main things:
- Using data from the database in the stage three viewers
- Serializing and restoring the student's app state between sessions

These items, particularly the second, aren't 100% done, but I want to get some of these updates into master before this branch falls too far behind. There's a lot of stuff here, but the main things are described below. 

## Specify student/class

The student and class used are specified via the `student_id` and `class_id` keyword arguments. If they aren't specified, the database won't be updated with any new data. For stage 3, one will get some "default" data from a particular (always the same) student and class from our generated data set.

## Stage 3 data

* The various pieces of data in stage 3, at least for the current setup of the viewers, are now all obtained from the database, thanks to some recent changes to API. Most of this data is queried during setup, but the data for the student's class is continually updated throughout the app runtime. I set the update frequency to be 30 seconds, but we can adjust this. Each time the update runs, the app queries the relevant data from the API, checks whether its data is out of date (using the `last_modified` field of the measurements), and updates the relevant glue data if necessary.

## State persistence

* During its setup, the app queries the API for the current student's state for the specified story, if one exists, and updates the local state accordingly. This is actually a relatively painless process - the glue state objects and Vue reactivity do a lot of work for us. In order to update each stage state, I've augmented our JSON encoder to be able to encode glue state objects, and added the stage state to the story state's `stages` dictionary.
* For MathJax inputs, we can now add a `tag` attribute to each `\input` command, and when MathJax typesets and create the input, any tagged inputs will automatically feed their current values into the story state (similarly, their values will be restored on startup). The syntax for adding a tag to the MathJax inputs is as follows:

`\input[id][classes][tag]{value}`

If no tag is specified, the input's value won't be retained. I've gone with a separate attribute from the input's `id` so that multiple inputs could relate to the same state value, if we want.

* To do this, I've added a custom HTML element class via the browser's `CustomElementRegistry`. This input has a static reference to the currently running app which allows this update to happen. Since custom elements are retained on a refresh, if this class already exists in the registry, the static variable which defines the app instance is updated. 

* For MC components, I defined an `mc-score` `CustomEvent`, which the main application listens for in order to update scoring. I went with a `CustomEvent` rather than a Vue event as Vue events don't bubble. Similar to the inputs, a `score-tag` prop will, if set, make the MC's score be retained in the state.

There are still some things to do here:
 - Pat and I discussed keeping track of a total score from the MCs for each stage, and the story overall, which hasn't been implemented yet
 - The JavaScript part of the `app.vue` has gotten pretty bloated. I want to look into splitting out some of this functionality into mixins, so that each story could potentially pick and choose which functionality it needs
 - I suspect there will be a lot of kinks to work out with the state persistence and things displaying correctly


patudom:
@Carifio24, In Stage 2, the angular sizes are no longer being entered in the table, so I can't measure distances and plot my data in Stage 3. I think something funky must have happened to that with the merge. Would you mind taking a look, please? Thanks!

Carifio24:
@patudom I've just pushed a commit that should fix that issue

PR #128, opened 2022-07-18
Allow enabling/disabling of tool buttons

Carifio24:
This PR adds functionality to our glue toolbar to allow enabling/disabling of tools (e.g. when a student reaches a certain point in the story). This is done by adding a `disabled` field to each tool's entry in the toolbar's `tools_data`, which we can then update in the appropriate way so that `ipyvuetify` responds to the change. To facilitate these updates, I've added `set_tool_enabled` and `is_tool_enabled` methods to the toolbar, which both identify tools by their IDs. For example, to disable the rest wavelength tool on the spectrum viewer in stage one, you would do

```
spectrum_viewer.toolbar.set_tool_enabled("hubble:restwave", False)
```

Calling `spectrum_viewer.toolbar.is_tool_enabled("hubble:restwave")` will return whether the tool is currently enabled or not.

### DIFF
--- a/cosmicds/components/toolbar/toolbar.py
+++ b/cosmicds/components/toolbar/toolbar.py
@@ -88,6 +88,9 @@ class Toolbar(VuetifyTemplate):
         tool = self.tools[tool_id]
         self.update_tools_data(tool, { "disabled": not enabled })
 
+    def is_tool_enabled(self, tool_id):
+        return not self.tools_data[tool_id]["disabled"]
+
     def add_tool(self, tool, data={}):
         self.tools[tool.tool_id] = tool
         self.update_tools_data(tool, data)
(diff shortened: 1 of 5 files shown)


Carifio24:
If anyone wants to test this out easily, open up the app in a Jupyter notebook, then add a cell with something like the following. You can change `tool_id` to test out other tools, or use a different viewer.

```
# The spectrum viewer is the first viewer that appears in the story
specview = app.session.application.viewers[0]
tool_id = "hubble:wavezoom" # The ID of the zoom tool

sv.toolbar.set_tool_enabled(tool_id, False) # or True
```

If you progress through the story to the point where the spectrum viewer is visible, and then run the code above, you should see the appropriate change in the spectrum viewer toolbar.

patudom:
@Carifio24, I added some code to turn a few spectrum viewer tools off/on at different points. I'm sure it can be done more elegantly than this, so please feel free to update what I've done, or merge if you think it's ok.

Carifio24:
@patudom I simplified the code very slightly, but the functionality is the same. I mostly removed the checks on whether the tool is enabled - there's no issue with setting the tool to be enabled even if it already is.

PR #129, opened 2022-07-19
Stage two formatting 3

heywooddogwood:
An assortment of updates, primarily to stage 2, but including stage 1 and overall app layout.

- remove focus/hover effect from measure button
- update app-level top-right icons -- including addition of running score
- make all WWT widget buttons green
- make sure of row selections where necessary
- countdown for ang_size measurements
- make the ang_size slideshow a stage unto itself
- reformat the scoring for multiple choice
- clean up text/format for nav drawer
- fix issue of a disappearing guideline during initial galaxy selection


patudom:
The angular size slideshow that transitions from Stage 1 to Stage 2 is broken and won't advance past Slide 2. Probably something to do with the merge we did in fd127c0. Will go ahead and merge now and fix this issue later.

PR #130, opened 2022-07-21
Spectrum viewer updates

Carifio24:
This PR makes some improvements to the spectrum viewer:
* The element tick marks are now taller, to make it clear that they aren't data points
* When the student clicks to select their wavelength, a gray line is left behind indicating their measured value. If a galaxy is selected that already has a measured wavelength, the gray line is placed at the location of the previous measurement.

Any suggestions on the visual aspect of this are welcome. Currently it looks like the screenshot below:
spectrum_viewer_updates


Carifio24:
@patudom I've added a white background to the user's wavelength label in the spectrum tool. AFAICT there isn't an actual way to set a background directly, but I was able to fake it with some clever mark positioning and ordering. In the course of doing this, I also learned that the "extra" space above/below the vertical axis bounds in the viewer is 10px on both top and bottom.

If we want the same thing for the gray label, I can add that in as well.

PR #131, opened 2022-07-25
Prevent measuring line from vanishing when one end is grabbed

Carifio24:
This is a small PR to fix the bug that causes the measuring line to vanish when one end is grabbed, before the user moves the line again.


patudom:
The data update is not related to this branch, but it seemed an easy thing to tack on to this small PR.

PR #132, opened 2022-07-27
All stages formatting 1

heywooddogwood:
loading as draft so that Pat can check why ellipticals are not showing up (or at least without color) in the stage 3 scatterplot


Carifio24:
@heywooddogwood If you change the color of the ellipticals (defined [here](https://github.com/cosmicds/cosmicds/blob/main/cosmicds/stories/hubbles_law/stages/stage_three.py#L197)), does that have any effect on the visibility of the layer?

If not, can you trying opening the app in a notebook, and then running the following in a new cell and tell me the output?
```
viewer = app.story_state.viewers()[9]
[(layer.state.layer.label, layer.state.alpha, layer.state.color) for layer in viewer.layers]
```

heywooddogwood:
Hey @Carifio24 , here is what I'm getting as the readout:

```
[('Elliptical', 0.5, 'blue'),
 ('Spiral', 0.5, 'green'),
 ('Irregular', 0.5, 'red')]
```

And here is a screenshot of the visual output:

Screen Shot 2022-07-27 at 3 03 28 PM

PR #133, opened 2022-07-28
Spectrum viewer updates, part 2

Carifio24:
This PR makes some additional visual updates to the spectrum viewer, which are outlined below. Suggestions on tweaks to these updates are welcome.

* The previous line label now has a white background as well
* The opacity for both label backgrounds has been set to 0.7, so that the data can be somewhat seen behind the background
* The positions of element labels have been adjusted so that the two labels don't overlap when the rest wavelength tool is active

I've also pinned `glue-jupyter >= 0.12` in `setup.cfg`. There was a bug in older versions of glupyter where all event callbacks were fired on every event, regardless of which events they were assigned to. I fixed this bug, and it will make things easier if we know we don't have to account for it. (This is relevant for some of the items in this PR, which rely on bqplot events to update their placement).

### DIFF
--- a/cosmicds/stories/hubbles_law/CosmicDS.ipynb
+++ b/cosmicds/stories/hubbles_law/CosmicDS.ipynb
@@ -36,7 +36,7 @@
     "\n",
     "from cosmicds.app import Application\n",
     "\n",
-    "app = Application(story='hubbles_law', update_db=True)\n",
+    "app = Application(story='hubbles_law', update_db=False)\n",
     "app"
    ]
   },
--- a/cosmicds/stories/hubbles_law/tools/rest_wavelength_tool.py
+++ b/cosmicds/stories/hubbles_law/tools/rest_wavelength_tool.py
@@ -78,5 +78,5 @@ class RestWavelengthTool(CheckableTool):
         ymin, ymax = scale.min, scale.max
         if ymin is None or ymax is None:
             return
-        self.line.y = [ymin, ymax * 0.93]
-        self.label.y = [ymax * 0.96]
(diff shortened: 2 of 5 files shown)


patudom:
@Carifio24 It's really cool that the gray line marker stays in place when you zoom in and out! But we noticed there is some funky behavior with the white background for the labels when you zoom in and out. The white background for the gray line stays either too wide or too narrow when transitioning into or out of a zoom, and there is a ghost of a white background from the blue measurement tool that gets left behind.

Screen Shot 2022-07-28 at 1 52 50 PM
Screen Shot 2022-07-28 at 1 52 08 PM
Screen Shot 2022-07-28 at 1 51 24 PM

Carifio24:
@patudom Ah yeah, I forgot to think about zooming with the label backgrounds. We already keep track of the conversion between the scale and the pixels inside the viewer, so I should be able to fix that pretty easily.

Carifio24:
@patudom I've made some changes that should fix the zoom and background issues. It's working well on my end, but let me know if there are any issues for you.

patudom:
The gray marker line persists if you switch to a new galaxy and then zoom in on a range that includes the marked wavelength of the previous galaxy.
Screen Shot 2022-08-01 at 10 49 12 AM

patudom:
Also, would it be possible to leave the gray marker in place after the zoom in tool is selected? It might feel confusing to the user that it disappears.

Carifio24:
@patudom Sure! I can change the gray line to still show with the zoom tool. I actually think that change will also solve the other issue with the line showing when it shouldn't. Should be a quick fix, I'll try and push a commit later.

PR #136, opened 2022-08-10
Removing stories from main application

nmearl:
This PR removes the individual stories from the main application and setups a plugin system that is checked upon initialization for any stories installed by the user. See https://github.com/cosmicds/cosmicds-hubble for the Hubble's Law story repo.


Carifio24:
@nmearl For testing this, I assume that you'd recommend installing this branch together with the cosmicds-hubble master in the same environment?

nmearl:
Yeah, that's right.

Carifio24:
Does it make sense to change `setup.cfg` in `cosmicds-hubble` to point here (just while testing), rather than the cosmicds master? I'm just thinking about checking if there are any issues with the removed dependencies that we were talking about earlier.

PR #138, opened 2022-08-16
Update tickmarks based on viewer/application state

Carifio24:
This PR adds functionality for updating the tick mark settings of a viewer automatically during runtime. All of the changes described here are added to our `cds_viewer` class factory so that all our viewers will benefit.

The basic idea is that we add `nxticks` and `nyticks` properties to the viewer state. These values serve as recommendations for how many ticks we want to display on each axis. Whenever the axis bounds are changed, a new tick spacing is chosen from a set of "nice" values (see [here](https://github.com/Carifio24/cosmicds/blob/update-ticks/cosmicds/viewers/cds_viewers.py#L18)) based on the axis range and appropriate tick number recommendation.

This PR also adds the (Vue-exposed) method `vue_set_viewer_nticks` method to the `Stage` class to allow stages to set the tick numbers from JavaScript as they see fit. Additionally, when a viewer is added to a stage via `add_viewer`, whatever name is used for the viewer will be added as a class to the root element of the `ViewerLayout` template, so that stages can find these viewers in the DOM more easily. The PR https://github.com/cosmicds/cosmicds-hubble/pull/3 does this tick number updating for the Hubble data story.

### DIFF
--- a/cosmicds/components/viewer_layout/viewer_layout.py
+++ b/cosmicds/components/viewer_layout/viewer_layout.py
@@ -1,6 +1,6 @@
 from ipyvuetify import VuetifyTemplate
 from ipywidgets import widget_serialization
-from traitlets import Bool, Dict, Instance, Unicode
+from traitlets import Bool, Dict, Instance, List, Unicode
 from ipywidgets import DOMWidget
 
 from ...utils import load_template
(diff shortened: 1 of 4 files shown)


patudom:
Thanks for this @Carifio24! I'm guessing this is functionality that should be brought over to glupyter as well?

Carifio24:
Yeah, this is certainly something that could be useful in glupyter, and I don't see anything in the implementation that would prevent this from being easily ported over.

PR #139, opened 2022-08-16
Keep track of overall multiple choice score

Carifio24:
This PR adds functionality for multiple choice components to interact with the application and story state, allowing us to keep track of the student's overall score, both at the story and stage level.

This is accomplished via browser events. The setup is as follows:
* When the MC is created, it emits a `mc-initialize` event, with its score tag included in the message's detail. The app listens for this event and sends a `mc-initialize-response` event, with the appropriate score value if found, or a message indicating no score yet exists if not
* When the correct answer is chosen for the MC, a `mc-score` event is emitted, whose details include the score, number of tries, and choice selected (in case there are multiple correct answers). The app listens for these events, and upon receiving one, updates the story's `mc_scoring` and `total_score` attributes appropriately.

As currently constructed, the results in `mc_scoring` are stored in sub-dictionaries keyed by stage. This allows us to know which stage an MC score came from, which will allow us to do stage-specific things, like calculate the total score for a given stage, etc.


Carifio24:
@patudom I just pushed a fix for the score calculation issue that we saw yesterday. Could you check whether it fixes the issue for you?

patudom:
Looks like everything is working correctly. I added score tags to the stage 2 intro slideshow in https://github.com/cosmicds/hubbleds/pull/8 and those are being correctly added to the tally also.

PR #140, opened 2022-08-16
Remove now-unneeded Hubble's Law content

Carifio24:
There are a few items left over in the `cosmicds/stories/hubbles_law` directory. My guess is that these files were created after whatever point the migration branch was created from main, and thus there was nothing in the git history to delete them, and so they persisted post-migration. This PR removes these leftover files.


nmearl:
Yeah, I noticed this recently as well. This PR looks good to me!

nmearl (approved):

PR #144, opened 2022-08-31
Hide bqplot legend "checkmark" symbols

Carifio24:
This PR hides the "checkmark" symbols that appear in the bqplot legends. I couldn't find a way to do this in bqplot itself, so this is accomplished using some CSS. Currently this is done at the top level of the app, and so these symbols will be hidden by default. In the unlikely event that we do want these a story at some point, the stages there can override this CSS.

### DIFF
--- a/cosmicds/app.vue
+++ b/cosmicds/app.vue
@@ -575,4 +575,7 @@ input {
   z-index: 40;
 }
 
+.g_legend .line {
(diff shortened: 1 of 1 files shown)


patudom:
Looks good - thanks! Just as an FYI for others, I had to recompile to make this update work.

PR #145, opened 2022-08-31
State testing

Carifio24:
This branch is for testing out how well our persisted state can accurately reproduce the UI. I've set up a [wiki page](https://github.com/cosmicds/cosmicds/wiki/State-testing-branch) with instructions on how to use this branch, as well as [an issue](https://github.com/cosmicds/cosmicds/issues/146) for logging bugs. This description will be updated once those are both created.

Currently this branch only has one commit, so I'll probably rebase once #139 is merged in, since the overall score value that's displayed is definitely a UI piece that we should be testing.


patudom:
Closing this branch now that we have added state testing functionality to main.

PR #147, opened 2022-09-08
Set visual properties of subset groups on creation

Carifio24:
Checking in from abroad with a quick update. This PR modifies the places where our code creates subset groups to use the functionality that I added (with this project in mind) in https://github.com/glue-viz/glue/pull/2297. This lets us specify the visual attributes of the subset group when it's created, rather than waiting until the subset group is created and then updating it. In order to use this, we need to pin to `glue >= 1.4`, so this PR does that. This should prevent any quick flashes of color-changing when the subset group is first created.

If someone's looking to test this out, the table <--> scatter plot interaction near the beginning of stage 3 is one area where we create a subset group (which component creates the group depends on which one the student uses first).

### DIFF
--- a/cosmicds/components/table/table.py
+++ b/cosmicds/components/table/table.py
@@ -171,8 +171,7 @@ class Table(VuetifyTemplate, HubListener):
     def _new_subset(self):
         state = self.subset_state_from_selected(self.selected)
         if self.use_subset_group:
-            subset = self.data_collection.new_subset_group(label=self._subset_label, subset_state=state)
-            subset.style.color = self.subset_color
+            subset = self.data_collection.new_subset_group(label=self._subset_label, subset_state=state, color=self.subset_color)
         else:
             subset = self._glue_data.new_subset(label=self._subset_label, subset=state, color=self.subset_color)
         if self._on_create_subset is not None:
--- a/cosmicds/message.py
+++ b/cosmicds/message.py
@@ -1,0 +1,7 @@
+from glue.core.message import Message
+
+__all__ = ["CDSLayersUpdatedMessage"]
+
+class CDSLayersUpdatedMessage(Message):
+    def __init__(self, sender, tag=None):
+        super().__init__(sender, tag=tag)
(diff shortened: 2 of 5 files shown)


Carifio24:
@patudom Here's an example of the color flashing that can currently happen.


https://user-images.githubusercontent.com/14281631/190702569-9f21bdc8-ccce-4eb2-bdae-1acf8f0903d0.mp4

patudom:
Thanks, @Carifio24. I was able to reproduce the glitch in main and see that it does not happen in this branch.

The behavior of the displayed fit lines may need tweaking on this branch. (If you prefer to fix it in a new branch, you can go ahead and merge this one in.) In main, if you display the fit line for the dataset, when you select a subset point from the table, it automatically displays the line for the subset also. 

Screen Shot 2022-09-19 at 11 54 20 PM

In this branch, it looks like it's drawing a subset line on top of the original line for the sample instead of the correct line for the subset. 

Screen Shot 2022-09-19 at 11 55 58 PM

You have to hide the lines and redisplay them for it to show up correctly.  (But if the blue subset line is already displayed, it will update correctly as you add/subtract data points to the subset).

Carifio24:
Okay, so I think I've got this sorted out - this ended up being a bit trickier than I expected. I think what was happening is that creating the subset and setting its visual attributes all in one got rid of the color flash, since the subset isn't created with some other color before being updated (which is good!). However, this means that a `SubsetUpdateMessage` from updating the style (which we listen for [here](https://github.com/cosmicds/cosmicds/blob/main/cosmicds/tools/line_fit_tool.py#L41). We only get a `SubsetCreateMessage` (listened for on the line above), but it seems that the layer hasn't been added to the viewer yet when we get this message, and so calling the line fit tool's `refresh` method doesn't actually update anything.

My initial thought to fix this was to use a `LayerArtistUpdateMessage`, but when this message is broadcast, its `layer_artist` member value is actually the layer state (which, at least to me, is very confusing and something I'll probably look at). The layer state isn't good enough for us - we need the actual layer artist since we use its `scales` to create the bqplot Line mark.

Thus, I've added a `CDSLayersUpdatedMessage` which is basically a wrapper around the `on_changed` method of the viewer's layer artist container. The line fit tool now listens for this message to detect when a layer is added to the viewer. Note that we still need the `SubsetUpdateMessage` to detect when the subset state (i.e. which points are selected) changes.

PR #159, opened 2022-10-18
Allow state testing on main branch

Carifio24:
This PR adds the keyword options used in the state testing branch, so that we can use the main branch for testing state reloading issues. This will depend on the changes in #158  which is why I've marked it as a draft.


patudom:
Merged this to #158 so we can test the changes implemented there.

PR #171, opened 2022-11-02
Html tooltip for table.vue

johnarban:
Adds HTML support in `table.vue` for the table headers, title, and tooltip

image


cosmicds/components/table/table.vue
```diff
@@ -37,7 +37,7 @@
             <v-icon>{{tool.icon}}</v-icon>
           </v-btn>
         </template>
-        {{tool.tooltip}}
+        <p style="max-width: 40ch;" v-html="tool.tooltip">  </p>
```
Carifio24, line 40: It doesn't affect the functionality, but I think we should aim to avoid inline CSS unless it really offers an advantage. Something like `class="tooltip"` and `.tooltip { max-width: 40ch }` in the `style` tag seems like a better approach to me.

johnarban, line 40: so perhaps just this
```suggestion
        <p class="tooltip" v-html="tool.tooltip">  </p>
```

PR #173, opened 2022-11-03
Allow using viewers in multiple stages

Carifio24:
This PR adds functionality to allow using the same viewer in multiple stages. In particular, a viewer created in one stage can now be accessed in a later stage.

As an example, if I added a viewer with the label `"my-viewer"` in stage one, I could now add it to stage two by simply calling `add_viewer(label="my-viewer")`. It's important to note that these two widgets will then represent the same glue viewer instance - any changes (zooming, adding layers, etc) that a student makes in one widget will happen in both.

Behind the scenes, this is done by creating a `viewers` dictionary in the story state, which we use to hold a reference to every glue viewer added. When `add_viewer` is called in a stage, it first check to see if a viewer with the same label has already been added - if so, it uses that rather than create a new viewer. To ensure that this works properly, I've also made sure to sort the keys of `story_entry` during setup in the registry (so we know that the stages are set up in order).

I've created a [`share-viewers` branch](https://github.com/Carifio24/hubbleds/tree/share-viewers) for testing this out. It has a dummy extra stage which adds the layer viewer from stage three using this functionality.

### DIFF
--- a/cosmicds/phases.py
+++ b/cosmicds/phases.py
@@ -48,6 +48,7 @@ class Story(CDSState, HubMixin):
     total_score = CallbackProperty(0)
     has_scoring = CallbackProperty(True)
     responses = DictCallbackProperty()
+    viewers = DictCallbackProperty()
 
     def __init__(self, session, *args, **kwargs):
         super().__init__(*args, **kwargs)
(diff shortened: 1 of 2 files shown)


johnarban:
tested with the [share-viewers](https://github.com/Carifio24/hubbleds/tree/share-viewers) and main branches.

PR #175, opened 2022-11-14
Sample data generation

Carifio24:
This branch is for generating measurements of the "sample" galaxy that each student will measure (spec-1572-53177-0186). This is based off the data generation branch, but here there's no random selection of galaxies - the sample galaxy is automatically selected.

If you specify a student ID using the `student_id` keyword argument, you'll make a sample measurement for that student. Otherwise, if `student_id` is omitted, a new seed student is created. Data is automatically submitted to the database and is stored in a table specifically for these sample measurements.


Carifio24:
@patudom I've updated this to account for the fact that we now want two sample measurements for each student. There are now two spectrum viewers in stage one, and two measurement tools in stage two. In each stage, the first viewer/tool corresponds to the first measurement, and the second to the second measurement. Which measurement is first/second is labeled in the table.

Everything about keyword arguments from the original PR description is still valid. If you or @johnarban want to try this out and do a few test samples, I can then check the database and confirm that everything looks okay.

patudom:
Closing because we are done generating seed data.

Issue #176, opened 2022-11-14
Component state not updating


Carifio24:
So while working on something earlier, I noticed an issue with the stage of the guideline components in stage one not updating. See the video below - the app doesn't move to the second selection guideline when I select a galaxy, and the third guideline still thinks I need 5 galaxies when I've already added one:

https://user-images.githubusercontent.com/14281631/201761078-90c97181-1c4f-4ae4-8baf-98b7a36ba06d.mp4

I tried restarting what I was doing, starting with a fresh environment, etc, but this issue continued to persist in both the notebook and voila, which was pretty puzzling since this worked perfectly fine in the beta and we haven't made any real changes to this stage since then. Note that changes to state controlled directly by the stage template itself seem fine, but the changes controlled in component templates aren't working.

After a bit of investigation, it seems to me that this issue pops up with `glue_jupyter >= 0.14.1`. Downgrading to 0.14.0 makes everything work fine again. Looking at the changes from that release, I think the particular update that's affecting this is https://github.com/glue-viz/glue-jupyter/pull/325.

I'm not sure exactly why this breaks our updating. The linked PR prevents sending an update if the serialized JSON of the state hasn't changed. Maybe this works fine for updating the stage itself, but necessary subsequent updates needed for component frontends aren't getting sent? The other question is whether there's something in our usage pattern that's bad and needs updating, or whether we just want to work around this.

Anyways, the easiest things to do until we get to the bottom of this are:
1. Pin `glue_jupyter <= 0.14.0`. This is quick and easy, but very obviously not a long-term solution, since this is one of our most important dependencies and we'll want future updates.
2. Create our own subclass of `GlueState` and override `on_state_change` to use the old implementation there.

@nmearl Would be curious if you have any thoughts on this

@patudom @johnarban Which version of glue-jupyter are you using in your environment? Have either of you noticed this issue?

Carifio24:
This was resolved with the new state handling approach introduced in https://github.com/cosmicds/hubbleds/pull/110 and fully implemented in later PRs (e.g. https://github.com/cosmicds/hubbleds/pull/113, https://github.com/cosmicds/hubbleds/pull/117).

PR #180, opened 2022-11-22
Add CDSGlueState to wrap glue state objects

Carifio24:
This PR adds a `CDSGlueState` wrapper class for the `GlueState` traitlet. The main use for this is to override the `on_state_change` method to match the previous implementation in `GlueState`. This serves as an implementation of option 2 in #176. If things change on that front, we can re-evaluate whether we need this or not.

### DIFF
--- a/cosmicds/app.py
+++ b/cosmicds/app.py
@@ -8,7 +8,7 @@ from echo import add_callback, CallbackProperty
 from glue.core import HubListener
 from glue.core.state_objects import State
 from glue_jupyter.app import JupyterApplication
-from glue_jupyter.state_traitlets_helpers import GlueState
+from cosmicds.cds_glue_state import CDSGlueState
 from ipyvuetify import VuetifyTemplate
 from ipywidgets import widget_serialization
 from traitlets import Dict, Bool, Int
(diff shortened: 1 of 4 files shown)


johnarban:
So, this doesn't break anything if merged. 

Perhaps (and I am thinking about [hubbleds:*Use CDSGlueState #105*](https://github.com/cosmicds/hubbleds/pull/105)) it would be good to just do the import as
```python
from cosmicds.cds_glue_state import CDSGlueState as GlueState
```
so that it is easy to switch back if we find a better way to fix issue #176 and want to return to using the default `GlueState`. Although maybe wrapping in this way is bad practice 🤷🏾 ?

Carifio24:
Personally I'm in favor of being more explicit. I can see the merits of both, but in this case we'd have to run through all of the files anyway to change the imports if/when we change this back.

We could also change this back the opposite way if this becomes unnecessary - i.e. change `cds_glue_state.py` to be just `import GlueState as CDSGlueState`. That way we don't need to touch either the imports or the component files, and we keep everything using an (explicit) extra layer of indirection, even if we don't actually do anything with it.

@nmearl do you have any preference on this?

Carifio24:
Closed this as https://github.com/cosmicds/hubbleds/pull/110 will provide the correct resolution to #176.

PR #182, opened 2022-12-05
Add dotplot viewer

Carifio24:
This PR adds a basic dot plot viewer for drawing dot plots. This is an adapted version of the histogram viewer with a few extra wrinkles to make things work.

A few notes:
* The layer artist of this viewer is a subclass of the bqplot histogram layer artist. The main differences are that we overwrite the constructor and `_scale_histogram` methods to account for the fact that we're using the histogram values to construct `ScatterGL` rather than `Bar` marks. Other than the different mark logic, these implementations are largely the same as the superclass ones.
  - If there's interest on the glue side, I think this is something that could be added as an option in the glue-jupyter histogram viewer. If that ends up being the case, this would just be functionality added to the existing layer artist, rather than requiring a subclass. But that's something we can think about later.
* The main challenge here was getting the dots to draw nicely. bqplot allows setting individual sizes, or one default size, on a `Scatter` mark. Since we want every dot the same size, this implementation sets the `default_size` attribute.
  - `default_size` is specified in pixels, and in particular it specifies the area of the mark. Since we're sizing our circles based on the height of the viewer, the method that I've used to set the size is
    * Find the diameter d, in pixels, that we want by dividing the relevant figure height by our y range (y_max - y_min)
    * Convert this to an area via f * (pi/4) * (d - s)^2, where f and s are factors to add some spacing between dots (rather than have them be exactly tangent).
- Figure height was also a bit tricky. To make this work, we have a `viewer_height` member value in the viewer state, which by default is set to 410px, the default size of the relevant part of the figure. If we change the viewer height Python-side, by using `viewer.figure.layout.height = '<some-number>px`, our viewer state will listen to this update and the dots will be resized accordingly. In the unlikely case that we use some CSS or JS to change the figure height, we'll have to manually invoke this somewhere in the Python code.

Feel free to experiment with the dot sizing settings, which are set [here](https://github.com/Carifio24/cosmicds/blob/dotplot/cosmicds/viewers/dotplot/layer_artist.py#L49). I picked scaling and spacing values that I thought worked well, but I'm open to suggestions.

I've [uploaded a notebook](https://gist.github.com/Carifio24/551162e165bb72bfc3e34bdd410bd2f4) for experimenting with this, which also has an example of what a subset highlighting a particular student's data will look like (this student has two data points in the dataset). If you don't already have it, the necessary data file is in the linked gist as well.

### DIFF
--- a/cosmicds/viewers/dotplot/layer_artist.py
+++ b/cosmicds/viewers/dotplot/layer_artist.py
@@ -3,7 +3,7 @@ from glue_jupyter.bqplot.histogram.layer_artist import BqplotHistogramLayerArtis
 from glue_jupyter.link import dlink, link
 from bqplot import ScatterGL, Scatter
 from numpy import inf
-from math import floor
+from math import floor, pi
 from glue.utils import color2hex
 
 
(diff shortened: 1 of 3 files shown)


patudom:
Because of the different range scales that are relevant (~200,000 km/s with outliers vs. ~100s km/s zoomed in), it would be helpful if we can dynamically adjust the number of displayed bins based on how large the displayed range is. (It looks like it's currently set at ~15 bins on every zoom level). 

When the full range (with outliers) is displayed and there are only 15 bins, the outliers don't display anywhere close to their actual values, which will be confusing for students. 

Would the dynamic binning be hard to implement, @Carifio24? Thanks!

johnarban:
Merging this now. We can add additional functionality (selecting a measurement, highlighting the student measurement, etc) later.

Carifio24:
@patudom Adding dynamic binning should be straightforward - we can add something to [this method](https://github.com/cosmicds/cosmicds/blob/main/cosmicds/viewers/dotplot/state.py#L14) that sets `hist_n_bin` based on the data range and viewer bounds.

PR #184, opened 2022-12-12
create contrast-brightness-control component

johnarban:
adds control for the brightness and contrast of whatever is added to the optional slot, and also emits the filter to the parent, where it can be used with `:style=['newstyle']`.

This is applied to the distance tool in `hubbleds` PR [114](https://github.com/cosmicds/hubbleds/pull/114)


patudom:
This is great! Thanks @johnarban.

2 requests:

- Can you add a reset option to the interface? Either a button or an icon with a tooltip. We could use the same mdi-cached icon that we use everywhere else to indicate a reset.
- When the user clicks on a table row to select a new galaxy, can you reset the contrast/brightness levels?

Carifio24:
This is very cool! The brightness/contrast functionality is working perfectly for me. I do have a couple of comments:

* For consistency with the rest of the codebase, I think the component file should be 2-space indented, and the filename should be snake_case rather than PascalCase
* I'm wondering if it would be better to have this core component only be the element with the sliders themselves (that is, without the expansion panels, which would then go into the distance tool template in the Hubble story). This would give us more flexibility as far as where we can put this functionality - for example, maybe in another story we want this in a popover rather than an expandable element.

johnarban:
Yeah, I'll update the formatting - and I can separate out the sliders. I think it is a good idea, and I can make each one optional too

@patudom - it does reset if you touch the button, but I can add a separate reset to the end of the slider. As for resetting it, I will have to look in to how to do that, I think it should be possible.

johnarban:
I believe I made all the requested changes, each is mostly its own commit. 

To do the reset, here I just added a `reset` prop that triggers a `reset_styles` function when it changes and is `True`. This should make it fairly generic.

Carifio24:
So I 100% understand why you did the listener on a prop, but I think it's worth mentioning that ipyvuetify actually offers a (kinda hidden) way to call JS methods from Python. This is actually used for example [here](https://github.com/cosmicds/hubbleds/blob/0f2551cdbc80ab635bf7392ac983a5929d6b094d/src/hubbleds/components/distance_tool/distance_tool.py#L61) to reset the canvas in the distance tool.

The idea is similar to how we call Vue methods from Python. Any method starting with `jupyter_` is exposed to Python, and can be called using the syntax in the linked example, with `jupyter_` removed from the name. So inside the distance tool template you could add a template ref (e.g. `ref="bc"`) to the brightness/contrast control component. Then wrap its reset method via a Jupyter-callable method in the distance tool's template:

```
methods: {
  ...
  jupyter_reset_style: function() {
    this.$refs.bc.reset_style();
  }
}
```

and then we can just call `self.send({"method": "reset_style", "args": []})` in the Python.

I'm not sure if this is a better approach or not, just wanted to throw it out there as an option.

patudom:
@johnarban, this is fantastic!! Thank you. Everything behaves as I would like it to. I'll check with Harry and Anna about stylistic updates.

I will leave it to you and @Carifio24 to decide which version you want to use from a technical point of view. Once you've sorted it out, feel free to merge.

PR #187, opened 2022-12-19
Speech synthesis for v-icons and MathJax

Carifio24:
This PR updates the speech synthesis component to handle our icons as well as MathJax. For both cases, the basic idea is the same: for each element that get matched by the selectors, we deep-clone that element and replace any icons or MathJax with a `text` element with the text that we want to use. Then we can just grab the `textContent` of the cloned element, like we were doing before.

For icons, we use that fact that `<v-icon>some-icon-name</v-icon>` ends up in the DOM as `<i class="... some-icon-name ..."></i>` (or `<button class="... some-icon-name"></button>` for an icon button). Since we're using MDI icons, this means that we can find the icon name by looking for a class that starts with "mdi-". We can then use the rest of the class name, with hyphens replaced by spaced, as the relevant text.

MathJax is a bit trickier. MathJax _can_ automatically generate speech strings for its output, which is then stored in the `aria-label` attribute of one of the `mjx-container`. From what I've seen, these speech strings aren't perfect, but they should get the job done as this isn't something we have the bandwidth to do ourselves. The issue is that a user has to opt-in (via a context menu that can be shown by right-clicking any MathJax) to having the speech text be generated. I've tried a few things but can't get this option to automatically be activated ([relevant SO question](https://stackoverflow.com/questions/60688026/how-to-activate-mathjax-accessibility-by-default)). Worst case, we can have a small dialog showing a user how to do this.

Since MathJax is generally something we want to read, I've also added a new default selector for that.

### DIFF
--- a/cosmicds/app.vue
+++ b/cosmicds/app.vue
@@ -348,8 +348,16 @@ export default {
         }
       },
       options: {
+        menuOptions: {
+          settings: {
+            assistiveMml: true
+          }
+        },
         a11y: {
-          speech: true
+          speech: true,
+          sre: {
+            speech: 'deep'
+          }
         }
       }
     };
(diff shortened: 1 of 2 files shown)


Carifio24:
@patudom I've added the ability to map specific icon names like we discussed yesterday. If the icon class is, say `mdi-some-name`, the entry in the map should have the form `'some-name': 'text to say'`. This map is stored as `iconNameMap` in the speech synthesis component.

PR #188, opened 2022-12-21
Disable headers by default

Carifio24:
This PR modifies the `scaffold-alert` component to allow making the header optional. The header will now only be shown if text is passed into the `headerText` prop. In the case that the header is not visible, the speech button is now moved to the bottom row. If any of @heywooddogwood @johnarban @patudom have ideas on a better place to put it, feel free to modify.

Currently all of our guidelines do have values for `headerText`, but I'll modify e.g. https://github.com/cosmicds/hubbleds/pull/116 to make that not be the case.


Carifio24:
@patudom I've resolved the merge conflict. Since the header text itself was made a computed property in #186, I removed `showHeader` here and have the template just check the truthiness of the header text.

PR #192, opened 2023-02-08
Color updates 2

heywooddogwood:
- set z-index for nav titles so that they stand out as white text.
- if mc-radiogroup is in a guideline, give it the answer/feedback v-alert a mostly black background.
-----
Navigation drawer updates:
- Avatar is now 'warning' red.
- Make text of navigation stages white/black.
- Add a black line to the bottom of headers (to help distinguish when dialogue content has scroll overflow.
-----
Change app bar src image (Still room to adjust, but the new image has less activity, so perhaps simplifies the design.)
-----
- darken the background of the piggy bank chip
- add divider to scaffolds
- turn speech-synthesizer into icon/button
- make text of v-card-actions white/black
- remove background from v-card-actions

### DIFF
--- a/cosmicds/app.vue
+++ b/cosmicds/app.vue
@@ -89,7 +89,7 @@
               class="mr-3"
             >
               <v-avatar
-                color="info lighten-1"
+                color="warning"
               >
                 <v-icon dark>mdi-account-circle</v-icon>
               </v-avatar>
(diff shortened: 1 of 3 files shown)


heywooddogwood:
This PR is tied to https://github.com/cosmicds/hubbleds/pull/122 in the hubbleds repository.

PR #195, opened 2023-03-16
Speech settings and improved speech behavior

Carifio24:
This PR makes some updates to the speech synthesis functionality. In particular:
* There's a new speech settings widget that allows students to set various options for the speech reading, as well as whether or not they want the text to automatically be read. This widget can be opened using the new speech icon in the app bar. These options are persisted in a "student options" table in the backend database.

speech_settings

* Additionally, there have been some new props added to the speech synthesizer to allow speaking to be properly updated when closing or switching slides in one of the dialog slideshows.

I've marked this as a draft for now (and will do the same for the associated hubbleds PR). If there are any differences between desired behavior and what's implemented here, let me know and I'll change things accordingly.

### DIFF
--- a/cosmicds/app.py
+++ b/cosmicds/app.py
@@ -15,7 +15,7 @@ from traitlets import Dict, Bool, Int
 
 from .events import WriteToDatabaseMessage
 from .registries import story_registry
-from .utils import CDSJSONEncoder, load_template
+from .utils import CDSJSONEncoder, debounce, load_template
 
 v.theme.dark = True
 
@@ -210,11 +210,14 @@ class Application(VuetifyTemplate, HubListener):
             "value": value
         })
 
+    @debounce(2)
     def _speech_rate_changed(self, rate):
         self._student_option_changed('speech_rate', rate)
 
+    @debounce(2)
     def _speech_pitch_changed(self, pitch):
         self._student_option_changed('speech_pitch', pitch)
 
+    @debounce(2)
     def _speech_autoread_changed(self, autoread):
         self._student_option_changed('speech_autoread', autoread)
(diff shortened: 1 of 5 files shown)


patudom:
### Issues
- Something is cutting off the speech synthesizer.  It cuts off in a different place when different speeds are chosen. (This happened for me on Slide 2 of the intro slideshow, with the Aristotle image).
- The Intro slideshow kept reading after I clicked "Get Started" and the slideshow was closed. The intro slideshows seem to need different handling than the dialogs.

### New Feature Requests

If these are available on the user's computer, let's use the following as the default voices:
1. Google US English (we think this is available for everyone using Chrome)
2. Tessa (This is available on Safari & Firefox on a Mac)
3. We still need to check what's available in Windows - https://mdn.github.io/dom-examples/web-speech-api/speak-easy-synthesis/)

For the Apple voices, add the (language-country) info to our dropdown list:

Screen Shot 2023-03-17 at 12 10 08 PM

Carifio24:
It looks like Windows has three built-in voices - David, Mark, and Zira (Chrome shows these three, and Firefox additionally has "Desktop" versions of David and Zira that, at least from what I can tell, sound identical to the non-Desktop versions). Of these, I found David to be the most robotic, but none of them match up to the Google voices. I'll put Zira in the default list behind the Google US English (I don't think the relative order of Zira and Tessa matters).

Carifio24:
Surprisingly, Edge has easily the most robust selection of voices (there are over 300 - most of them must be bundled into Edge itself), and I would say the highest quality as well. I suspect there won't be a ton of students using Edge, but the ones who are should definitely get one of those (I'm leaning towards Microsoft Aria).

patudom:
Interesting. Yes, I would add a good Edge default, and I think any that are "good enough" will be just fine - no need to comb through all 300 of them. 😄

PR #199, opened 2023-03-28
Prevent line draw tool from getting stuck

Carifio24:
This PR aims to resolve https://github.com/cosmicds/hubbleds/issues/135. The behavior here is changed so that if the line mark has been created but the endpoint hasn't when the tool is deactivated, the line is removed (basically, the tool is reset). @patudom Let me know if we want a different behavior here.

Note that this doesn't need to be merged in before any stage renaming - that won't have any effect on this repo.


Carifio24:
This last commit, together with https://github.com/cosmicds/hubbleds/pull/138, should hopefully solve the tooltip issue.

PR #203, opened 2023-04-05
Fixes for DotPlotViewer to work with PR #202

johnarban:
This fixes an issue with dot plots that come up when the line mixin from @Carifio24's #202 is included. This includes the changes from #202 and should be merged after.

### DIFF
--- a/cosmicds/mixins.py
+++ b/cosmicds/mixins.py
@@ -68,7 +68,7 @@ class LineHoverViewerMixin:
 
         super().__init__(*args, **kwargs)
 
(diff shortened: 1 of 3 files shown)


Carifio24:
While this seems to work fine, I think that if there is a problem then we should fix it in the mixin - we don't want to have to make updates to any base class that we want to use this with, since in principle they should be entirely unaware of the mixin. I'll take another look at the mixin and see if I can resolve this.

johnarban:
Issue is fixed in #202

PR #212, opened 2023-04-14
add x range padding to dot plot viewer when resetting limits

patudom:


johnarban:
a simple enough change, it works with dotplot viewers when tested. .. merging

PR #214, opened 2023-04-20
Require free responses to be filled out to advance

Carifio24:
This PR adds functionality to the scaffold alert component so that free responses are required to be completed (currently defined as "the text field isn't empty"). This requirement is set to be true by default, but can be changed using the `require-fr` prop.

This is achieved by finding all of the free response components inside a guideline and, if there are any, checking whether each free response is complete whenever there's an input event. The nice thing about this approach is that we don't need to add anything to the actual guidelines themselves - everything is contained within the scaffold component.

Here's what this looks like in action:

https://user-images.githubusercontent.com/14281631/233387510-23a8fe53-4ffd-4c55-86c2-b38c44c4b8d3.mp4

### DIFF
--- a/cosmicds/components/free_response.vue
+++ b/cosmicds/components/free_response.vue
@@ -44,8 +44,7 @@ module.exports = {
   data: function () {
     return {
       response: "",
-      initialized: false,
-      filledOut: false
+      initialized: false
     };
   },
 
@@ -81,11 +80,6 @@ module.exports = {
     onBlur(_event) {
       if (this.tag !== undefined) {
         this.dispatchUpdateEvent();
-
-        // Normally we would just have this be a computed
-        // but we want this to be in sync with the response
-        // that's been emitted, which only happens here
-        this.filledOut = this.response.length > 0;
       }
     },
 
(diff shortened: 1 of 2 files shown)


johnarban:
So this seems to need to have a little bug when on `con_int2` `guideline-confidence-interval-reflect-2`, when you come to the question, the `next` button is active. But if I then go back and then return to the question the `next` button is properly inactive (see video). If instead of clicking `back`, I just `next` when I get to the next new guideline with a `free-response` that `next` is properly disabled.  This happens in the notebook and voila. If go straight to `con_int2c` (the one after `con_int2`) it loads fine 🤷🏾 

https://user-images.githubusercontent.com/7862929/233803299-400e1fed-1c5c-4576-a75c-2447b852b25e.mov

Carifio24:
Good catch! I'll look into that.

Carifio24:
So I've switched to using a `MutationObserver`-based approach that should fix the above issue. It should also better handle cases where we add (or remove) free responses (e.g. multi-part guidelines).

PR #215, opened 2023-04-21
fix x range padding to work with no layers visible in dotplot

johnarban:
this isn't a "real" interaction, so we don't want associated callbacks to be called.

This fixes a small issue introduced in #212 where changing `x_max` or `x_min` fires a callback to update the y limits while no layers are visible. This is only a problem when hubble ds is at https://github.com/cosmicds/cosmicds/commit/64cce84a63bbfcda14de0ecddc66c11187cc0d76 (main as of Apr 21) because this version has no layers loaded when `viewer.state.reset_layers()` is called. Later version have at least one visible layer.

### DIFF
--- a/cosmicds/viewers/dotplot/layer_artist.py
+++ b/cosmicds/viewers/dotplot/layer_artist.py
@@ -125,13 +125,13 @@ class BqplotDotPlotLayerArtist(BqplotHistogramLayerArtist):
         else:
             self.state._y_min = 0
 
-        largest_y_max = max(getattr(layer, '_y_max', 0)
-                            for layer in self._viewer_state.layers if layer.visible)
+        largest_y_max = max((getattr(layer, '_y_max', 0)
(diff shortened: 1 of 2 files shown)


Carifio24:
Just curious, is there a reason that we want to report pre-padded values to observers, instead of doing the padding inside `delay_callback`?

johnarban:
So that is where the padding was originally being added. the `delay_callback` only puts off the redrawing of the histogram until the block is complete. 

I think since `super().reset_limits()` has already been called, any callbacks associated with the `x_min` and `x_max` that we want to happen already have, and the padding is just cosmetic and shouldn't cause any other changes.

Carifio24:
Those callbacks shouldn't be triggered yet since the `super.reset_limits()` call happens inside the `delay_callback` context manager. If it's a problem with the case of no layers being visible, I feel like going with some default xmin/xmax instead makes more sense.

PR #217, opened 2023-04-26
Let viewers know their size

Carifio24:
This PR adds functionality that lets CosmicDS viewers be aware of their size on page. This is useful for things like the dotplot viewer (to which this PR applies this functionality, so this should help with https://github.com/cosmicds/hubbleds/issues/161), where the mark size depends on the figure size. The moving line functionality could also have used something like this - we were able to use pointer events to determine this information and it's not worth changing now, but this should help make similar things in the future easier.

### DIFF
--- a/cosmicds/components/viewer_layout/viewer_layout.py
+++ b/cosmicds/components/viewer_layout/viewer_layout.py
@@ -1,6 +1,6 @@
 from ipyvuetify import VuetifyTemplate
 from ipywidgets import widget_serialization
-from traitlets import Bool, Dict, Instance, List, Unicode
+from traitlets import Bool, Dict, Instance, Int, List, Unicode, observe
 from ipywidgets import DOMWidget
 
 from ...utils import load_template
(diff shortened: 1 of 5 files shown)


patudom:
Also, the marker size seems to change between guidelines even though the data are the same. Do we need to do something to initialize the marker sizes properly?

This is at Stage 3, dot_seq1:
Screen Shot 2023-04-26 at 7 31 40 PM

which is the same as the top viewer here at dot_seq4.

Screen Shot 2023-04-26 at 7 31 50 PM

Carifio24:
Hmmm...so I suspect the reason they aren't getting any larger is due to the number of bins. It looks like the bin count is set to 75 for the stage 1 dotplot viewers (and I don't see anywhere that it changes).

The updated size calculation does (viewer height / # bins) to get an estimate for the max allowed size from the x-direction. Since the number of bins (rather than the bin size) is constant when zooming, the x-bounds don't have an effect on the x portion of the calculation. It looks weird here because there are a lot of (empty) bins present.

patudom:
In the past, zooming in to a narrower x-range made the points larger, I guess because it was always recalculating the size based on the total number of points it had to plot in the y dimension. Is there a not-overly complicated way to know how many bins are in the zoomed in range and account for that? If not, we can leave it as is for now and refine after the May pilot.

Carifio24:
Yeah, the old calculation only took the vertical dimension into account. When the points got redistributed, this generally meant a decrease in `y_max` and thus larger points.

Now, to prevent the sort of horizontal overlapping we were seeing, I added in another check based on the viewer width and the number of bins. The number of bins stays constant until it gets manually changed - so if that's what setting the size, it likely won't change when zooming in (unless the vertical condition becomes more restrictive).

PR #222, opened 2023-04-30
Increase padding on dot plot viewer so the label on the measuring 

patudom:
tool can be read, else it sometimes falls off the edge of the viewer.
Screen Shot 2023-04-29 at 8 14 32 PM


johnarban:
Before padding
image

After
image

So it will still cutoff the label if you get too close to the edge, and adding the padding makes the axes labels over-crowded

patudom:
Closing for now because we think the poorly-spaced x-axis tick marks is worse than the moving label falling off the side of the graph.

patudom:
For the future, we do want to either switch the moving label to the left when it gets too close to the right edge of the graph, or space the x-axis tick marks better.

PR #224, opened 2023-05-03
Improvements to speech settings

Carifio24:
This PR makes the following changes to the app's speech settings:

* Only a finite set of voices are allowed to be selected. In general, this should allow 1-3 options for each student, depending on their browser and operating system. The allowed voices are those whose `voiceURI` value is included in the `voiceURIs` data value in the speech settings component.
* Some voice/browser combinations have rate adjustments based on the timing tests that I ran. The most significant of these is Google US English, both due to its behavior and because I expect it will be the most-used voice. In my testing, this voice seems to have a "true" speed rate which is a quadratic function of the rate assigned to the utterance; thus, we attempt to (roughly) undo that here.

### DIFF
--- a/cosmicds/components/speech_settings.vue
+++ b/cosmicds/components/speech_settings.vue
@@ -77,6 +77,35 @@
 
 <script>
 module.exports = {
-  props: ['state']
+  props: ['state'],
+  created() {
+    window.speechSynthesis.onvoiceschanged = (_event) => {
+      this.updateVoiceList();
+    };
+    this.updateVoiceList();
+  },
+  data() {
+    return {
+      voiceURIs: [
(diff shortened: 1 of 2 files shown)


patudom:
Works great in Chrome. Failed in Firefox with this error:
```
    TypeError: utterance.voice is null
    transformRate index.js:302
    makeUtterance index.js:148
    utterances index.js:243
    speak index.js:243
    click index.js:3
    VueJS 3
```

patudom:
Now runs without crashing on Firefox but the voice rate and timing needs adjusting after the pilot. Will tell pilot teachers to encourage students to use Chrome or don't use the voice reader.

PR #234, opened 2023-05-07
Layout updates and instructions snackbar

patudom:
- Add contain:layout to styling, which keeps the app all together in my notebook in Chrome (but doesn't work in every browser)
- Slightly reduce size of title bars, width of left menu, etc
- Remove logos in footer
- Add instructions snackbar that points out audio controls and light/dark toggle. 

Issues with snackbar:
- It probably doesn't need to appear after the first or second time a user has accessed the app and gets annoying in testing, but we can fix that later.
- I can't figure out how to get the snackbar text to be black in light mode. Ignoring for now.

### DIFF
--- a/cosmicds/app.vue
+++ b/cosmicds/app.vue
@@ -274,7 +274,7 @@
         top
         center
         color="info"
-        text="app_state.dark_mode ? 'white' : 'black!important'"
+        :text="app_state.dark_mode ? 'white' : 'black!important'"
         style="font-size: 1.2rem; line-height: 2"
         multi-line
         elevation="24"
--- a/cosmicds/app.vue
+++ b/cosmicds/app.vue
@@ -731,4 +731,8 @@ mjx-assistive-mml cds-input {
   margin: 2px;
 }
 
+#inspire{
+  contain:layout
+}
+
 </style>
--- a/cosmicds/app.vue
+++ b/cosmicds/app.vue
@@ -740,12 +740,12 @@ mjx-assistive-mml cds-input {
   contain:layout;
 }
 
-/*
-Don't think this is doing anything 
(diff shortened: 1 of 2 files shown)


cosmicds/app.vue
```diff
@@ -268,11 +268,44 @@
...
       </v-row>
+      <v-snackbar
+        v-model="show_snackbar"
+        timeout="200000"
+        transition="fab-transition"
+        top
+        center
+        color="info"
+        :text="app_state.dark_mode ? 'white' : 'black!important'"
+        style="font-size: 1.2rem; line-height: 2"
```
johnarban, line 287: I would suggest using `font-size: 1em; line-height: 1.5`

cosmicds/app.vue
```diff
@@ -268,11 +268,44 @@
...
+          </div>
         </v-col>
       </v-row>
+      <v-snackbar
+        v-model="show_snackbar"
+        timeout="200000"
+        transition="fab-transition"
+        top
+        center
+        color="info"
```
johnarban, line 285: I think this could just be the default color rather than the guideline color (so could remove `color="info"`

cosmicds/app.vue
```diff
@@ -53,7 +53,7 @@
                 v-on="on"
                 @click="speech_menu = !speech_menu"
               >
-                <v-icon>mdi-voice</v-icon>
+                <v-icon>mdi-tune</v-icon>
```
johnarban, line 56: Perhaps use [`mdi-tune-vertical`](https://pictogrammers.com/library/mdi/icon/tune-vertical/). Both here and L292

johnarban (approved):

PR #236, opened 2023-05-12
Allow dotplotviewer to optionally use default histogram bars

johnarban:
This is a small enhancement to make the dotplot viewer a bit more broadly useful. Basically giving a histogram with the unique features of the dotplot (rescaling bins and height). It is set up to that it doesn't change the current workflow. 

Normal operation: `HubbleDotPlotView()`
image

With ~`HubbleDotPlotView(use_bars = True)`~ `viewer.state.use_bars = True`
image

### DIFF
--- a/cosmicds/viewers/dotplot/layer_artist.py
+++ b/cosmicds/viewers/dotplot/layer_artist.py
@@ -71,6 +71,7 @@ class BqplotDotPlotLayerArtist(BqplotHistogramLayerArtist):
             self.bars = Bars(**style, x = [0,1], y = [0,1])
         else:
             self.bars = Scatter(**style, x = [0,1], y = [0,1], marker = 'ellipse', default_skew = 0)
+            dlink((self.state, 'skew'), (self.bars, 'default_skew'))
         with delay_callback(self._viewer_state, 'y_min', 'y_max'):
             self._scale_histogram()
         self.view.figure.marks = marks + [self.bars]
--- a/cosmicds/viewers/dotplot/layer_artist.py
+++ b/cosmicds/viewers/dotplot/layer_artist.py
@@ -27,7 +27,7 @@ class BqplotDotPlotLayerArtist(BqplotHistogramLayerArtist):
             'rotation': LinearScale(min=0, max=180)
         }
         
-        if getattr(self.view,'use_bars',False):
+        if self.view.state.use_bars:
             self.bars = Bars(
                 scales=self.view.scales, x=[0, 1], y=[0, 1])
         else:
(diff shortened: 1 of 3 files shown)


Carifio24:
Overall I'm a fan of the idea here. But `use_bars` feels like something that should be in the viewer state, not the viewer itself (and as a `CallbackProperty`, so the display can update dynamically).

johnarban:
That's a good idea. I have added to commits to implement your suggestion

cosmicds/viewers/dotplot/layer_artist.py
```diff
@@ -26,16 +26,23 @@ def __init__(self, view, viewer_state, layer_state=None, layer=None):
...
                             marker='ellipse')
+            
 
         self.view.figure.marks = list(self.view.figure.marks) + [self.bars]
 
         dlink((self.state, 'color'), (self.bars, 'colors'), lambda x: [color2hex(x)])
         dlink((self.state, 'alpha'), (self.bars, 'opacities'), lambda x: [x])
-        dlink((self.state, 'skew'), (self.bars, 'default_skew'))
+        if  not self.view.state.use_bars:
+            dlink((self.state, 'skew'), (self.bars, 'default_skew'))
```
Carifio24, line 45: This looks good here, but I think we want to make sure to handle this correctly in `_update_bar_type` as well.

cosmicds/viewers/dotplot/layer_artist.py
```diff
@@ -53,10 +60,24 @@ def __init__(self, view, viewer_state, layer_state=None, layer=None):
         add_callback(self._viewer_state, 'hist_n_bin', self._update_size)
         add_callback(self._viewer_state, 'viewer_height', self._update_size)
         add_callback(self._viewer_state, 'viewer_width', self._update_size)
```
Carifio24, line 62: This isn't a problem with this PR, but this currently doesn't quite work because `viewer_height` and `viewer_width` were removed from the dotplot viewer state when the support for viewers knowing their dimensions was added to `cds_viewer` (pretty sure this was my fault). It'll be fine with the Hubble dotplot viewer (which does get the `cds_viewer` wrapper), but we should add in `viewer_height` and `viewer_width` properties to the dotplot viewer state so that this works as a standalone thing.

cosmicds/viewers/dotplot/layer_artist.py
johnarban: Would the right raw to handle this be to remove the link and then recreate it? Is there a way to disable/enable a link in glue?

cosmicds/viewers/dotplot/layer_artist.py
johnarban: Alright, I've added those back here 907b00ae8dbb38430c363a7d89526af28f88b4b5

cosmicds/viewers/dotplot/layer_artist.py
johnarban: I just relink the skew paramter here 8335b1112be7a269e0a1e6c9981cd49fa43f46a6. I'm not sure if there is some cleanup of the old link that needs to happen, but it seems fine

johnarban:
with the above changes `viewer.state.use_bars = True` will turn it into a regular histogram

PR #243, opened 2023-07-12
Add percentage and statistic selectors

Carifio24:
This PR adds two components:

* One allows students to select a percentage (by default 50%, 68%, and 95%), and creates a subset that encompasses that percentage of the specified data/component around its median.
* The other allows students to select statistics (by default mean, median, and mode), and draws vertical lines on the given viewer at the locations of those statistics.

- [x] Update to handle multi-modal data

### DIFF
--- a/cosmicds/components/__init__.py
+++ b/cosmicds/components/__init__.py
@@ -1,4 +1,6 @@
 from .generic_state_component import *
+from .percentage_selector import *
+from .statistics_selector import *
 from .table import *
 from .toolbar import *
 from .viewer_layout import *
--- a/cosmicds/components/layer_toggle/layer_toggle.vue
+++ b/cosmicds/components/layer_toggle/layer_toggle.vue
@@ -39,8 +39,8 @@
 
 <style scoped>
 
-.v-card{
-  border: solid 1px black!important;
+.v-card {
+  border: solid 1px black !important;
 }
 
 </style>
--- a/cosmicds/components/percentage_selector/__init__.py
+++ b/cosmicds/components/percentage_selector/__init__.py
@@ -1,0 +1,1 @@
+from .percentage_selector import PercentageSelector
--- a/cosmicds/components/percentage_selector/percentage_selector.py
+++ b/cosmicds/components/percentage_selector/percentage_selector.py
@@ -2,7 +2,6 @@ from ipyvuetify import VuetifyTemplate
 from numpy import array, percentile
 from traitlets import Float, Int, List, Unicode, observe
 
-from glue.core import Data
 from glue.core.subset import RangeSubsetState
 
 from ...utils import load_template
@@ -10,11 +9,9 @@ from ...utils import load_template
 class PercentageSelector(VuetifyTemplate):
     
     template = load_template("percentage_selector.vue", __file__, traitlet=True).tag(sync=True)
-    color = Unicode("#1e90ff").tag(sync=True)
+    radio_color = Unicode("#1e90ff").tag(sync=True)
     options = List([50, 68, 95]).tag(sync=True)
-    selected = Int(allow_none=True).tag(sync=True)
-    selected_min = Float().tag(sync=True)
-    selected_max = Float().tag(sync=True)
+    selected = Int(None, allow_none=True).tag(sync=True)
     unit = Unicode().tag(sync=True)
     was_selected = Int(allow_none=True).tag(sync=True)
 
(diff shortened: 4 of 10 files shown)


patudom:
@Carifio24, this is working well - thanks so much for implementing these. 

Would it be possible to also:
- Allow the stories to specify a unit on the labels on the percentage ranges and the vertical x-value indicator value
- Allow user to unselect the radio button on the ranges.

Carifio24:
@patudom These two most recent commits (plus one on Hubble) add those features. Let me know if you see any issues

patudom:
@Carifio24, @heywooddogwood and I did quite a bit of testing on this excellent new functionality and we have the following requests. 

### Quick'ish (we hope) things:
- [ ] Please restyle the mean/median/mode check boxes to match the percentage radio buttons (and only allow one stat to be selected at once). (When you can select multiple stats at once, if they all the same value, the only one you see is the mode, which will confuse the user.  Harry and I pondered whether it would make sense to just give the most recently clicked stat line marker the highest z-index, but we think it will be more intuitive for the two new tools to have the same look and functionality. It's easy enough to toggle between stat buttons to compare them.)
- [ ] If a stat is already displayed, when you click a percentile range, the stat line marker goes away. Is it possible to keep the stat marker from going away? (If that's not possible, the stats buttons should be unselected when the lines disappear, else they end up out of sync and you have to click the button twice to get the stat marker back.)
- [ ] When you zoom in, the stat line marker for mean/median/mode needs to be redrawn in the new location. It currently remains static and does not line up with the new location of the marked value.
Screen Shot 2023-07-12 at 7 10 28 PM

- [ ] Harry suggested a color scheme for the percentage subsets that would be more intuitive for the user. Instead of turning the percentage subset range a new color, can you turn the original, full distribution gray, and give the new percentage subset the original color of the full distribution? (The idea is that we want to give students the sense that the outer x/2% of the data points are fading away. Also, we think it will be less overwhelming not to introduce lots of new colors that don't have particular significance). This is a crappy zoom sketch, but hopefully gives the idea.
Screen Shot 2023-07-12 at 7 08 49 PM

patudom:
This is more complicated, so we should discuss whether this would be manageable in a short timescale and how exactly we would want this to work.

- [ ]  How difficult would it be to have an implementation where 1 set of buttons displays the stats/percentage ranges on 2 different viewers? The last screenshot in the above comment demonstrates what we'd like, that if the student clicks "68%", it would display that percentage range in both graphs, so the students can see more intuitively that the bottom distribution covers a larger fraction of the data points within a narrower range. In this case, we would need to move the label with the range to the viewer.

patudom:
We noticed that the displayed range for this graph goes past where the last bin is, which doesn't seem to make sense. (The last bin is 21 Gyr, but the high end of the 95% range is given as 22 Gyr).
Screen Shot 2023-07-12 at 6 39 20 PM

I wondered if maybe the code here is not necessary? https://github.com/cosmicds/cosmicds/blob/05bfd20754ccea0d30d54aa267ba3067a965fe25/cosmicds/components/percentage_selector/percentage_selector.py#L68C1-L73C38

Is there a reason we aren't just reporting these values (rounded) as the relevant range? 
```      
bottom = percentile(data, bottom_percent, method="nearest")
top = percentile(data, top_percent, method="nearest")
```

patudom:
- [ ] For stats selector, add buttons (next to the stat name, maybe questions mark icons) that user can click to pop up a dialog box with a definition of the statistic and a supporting diagram.

patudom:
This is looking fantastic - thank you, @Carifio24! I noticed a couple small things:

- [ ] When you have a stat up and then click on a percentage range, the stat marker remains in the top viewer, but disappears in the 2nd viewer.
- [ ] On my computer, the range label is overlapping the stat legend. Is there a way to adjust the positions (and also make the text sizes more comparable to each other)?
Screen Shot 2023-07-13 at 5 19 56 PM

patudom:
- [ ] The upper/lower transforms still need tweaking. When you click "95%" for the all classes viewer, the bin at 12 Gyr is highlighted, but the lower end of the range is displayed as 13 Gyr, so I wonder if we actually wanted ceil for the lower transform too? 

Actually, here with the 50% ranges shown, the lowest displayed bin is 2 Gyrs smaller than the reported lower range in both the student and classes plots, so there may be something more complicated happening than just the 0.5 edge cases?
Screen Shot 2023-07-13 at 5 53 10 PM

patudom:
- [ ] I set up my class to have 2 modes, but only 1 is being displayed. 
- [ ] The unit label is now appearing in the label and I'm not sure why.
Screen Shot 2023-07-14 at 4 17 59 PM

patudom:
- [ ] It would be helpful to add ? definitions to the percentage ranges too (like the mean, median, mode). They can say basic things like “The range of values shown are the inner 50% of data points. This means that 25% of the data points in the distribution have values less than this range, and 25% of the data points have values greater than this range.”

patudom:
- [ ] Something funky is happening with the percentages for small numbers. I would expect the inner 50% of data for 6 students to include 3 or 4 values. This is displaying 5 values. 

Screen Shot 2023-07-15 at 11 28 59 AM

PR #249, opened 2023-07-17
Add validation functionality to free responses

Carifio24:
This PR adds the ability to do input validation to the free response component. This comes in two forms:
- Preset "types" that define common use cases (integers, floats, unsigned (non-negative) integers)
- Custom validation rules

Specifying the `type` prop of a free response will activate the relevant rules. This includes a type-specific error message as well (if a custom one isn't specified). Additionally, one can specify custom validation via `rules`, as well as a specific `helpMessage`. If one uses both `type` and `rules`, the rules will be applied on top of the given type.

Also, while by default the free responses require a response (they can't be empty), an `allowEmpty` prop allows that to be modified, so that we can have optional free responses.

### DIFF
--- a/cosmicds/components/free_response.vue
+++ b/cosmicds/components/free_response.vue
@@ -36,7 +36,7 @@ module.exports = {
     },
     helpMessage: {
       type: String,
-      default: "Invalid format"
+      required: false
     },
     hint: {
       type: String,
@@ -68,6 +68,7 @@ module.exports = {
     return {
       response: "",
       initialized: false,
+      defaultHelpMessage: "Invalid format",
 
       allowedInput: {
         int: {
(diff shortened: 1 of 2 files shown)


Carifio24:
@patudom I've added a sample implementation of this in one of the guidelines [here](https://github.com/Carifio24/hubbleds/blob/fr-rules/src/hubbleds/components/generic_state_components/stage_5/guideline_confidence_interval_reflect2_c.vue). If you're using this branch, use `go_to_marker("con_int2c", 5)` to get to the relevant guideline.

The rule requiring that one of the ages be greater than 3 doesn't make any sense in context, it's just a demo to show what we can do with this setup.

Issue #246, opened 2023-07-15
Lingering issues with percentage and stats selector functionality


patudom:
We are merging in #243 for now, but I'm noting here some issues that still remain with its functionality.

- [x] When you have a stat displayed and then click on a percentage range, the stat marker remains in the top viewer, but disappears in the 2nd viewer.
<img width="1315" alt="Screen Shot 2023-07-15 at 11 54 58 AM" src="https://github.com/cosmicds/cosmicds/assets/12750048/15693a7d-afe1-41a8-a3b0-f9a0772cac26">

- [x] The percentage selector is not handling small class size correctly. I would expect the inner 50% of data for 6 students to include 3 or 4 values. This is displaying 5 values.
<img width="634" alt="Screen Shot 2023-07-15 at 11 28 59 AM" src="https://github.com/cosmicds/cosmicds/assets/12750048/6fc3ad09-8af7-4254-8bc4-ac42274f24a1">

patudom:
- [ ] When a student clicks any ? to open a definition, it would make sense for that stat/range to be displayed if it is not already. Is that something that would be straightforward to add? (i.e., clicking ? activates the radio button if it isn't already active.)

johnarban:
One other issue: The highlighted range in 13-15, but the label says 14-15. It looks like only one thing is selected in this. and it just so happens that the highest value in that bin is 13.832..... (this is using student 4070)

<img width="738" alt="image" src="https://github.com/cosmicds/cosmicds/assets/7862929/5fe5da11-6a4a-4365-bad0-c5bdd141905d">

Carifio24:
There are fixes for these issues in #248. The only thing that I'm not 100% satisfied with is the percentage/figure title - I moved it over to the left a bit to try and avoid clashing with the legend, but I'm still looking for a robust solution.

In that PR, I changed the way that the percentile works. What I had been doing is using the `Data`'s built in percentile calculation to find the upper and lower percentiles, and then defining a range subset between these two values. But when one can have duplicate data values, this isn't the same thing as actually taking X% of the points, which is particularly noticeable when it pops up in smaller data sets. I've modified the percentile calculation to be based on sorting indices, meaning that we're actually taking X% of the data points (or as close to X% as the size of the data will allow).

[...]

PR #258, opened 2023-08-01
Updates to selector bounds

Carifio24:
This PR attempts to address the issues with the bounds from regions selected using the percentage selector (see the second item in https://github.com/cosmicds/cosmicds/issues/246#issuecomment-1640330084). The logic for the selector now takes into account how many items there should be in the percentage range, and makes sure that this is the case by either dropping the edge value further from the mean, or adding the new one closer to the mean, as necessary.

Also, some of the generic logic has been refactored out into utility functions so that it can be used in other places (e.g. calculating the relevant statistics for the student's particular class data in the Hubble story).

### DIFF
--- a/cosmicds/components/percentage_selector/percentage_selector.py
+++ b/cosmicds/components/percentage_selector/percentage_selector.py
@@ -1,5 +1,3 @@
-from math import ceil, floor
-
 from ipyvuetify import VuetifyTemplate
 from numpy import argsort, array
 from traitlets import Int, List, Unicode, observe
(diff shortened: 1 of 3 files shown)


patudom:
Looks like we're super close on this! We noticed a weird edge case with `student_id=4094` where the 95% for 7 values should be highlighting everything, but instead it's highlighting 6 values, and the one it's missing is farther away from the median.
Screen Shot 2023-08-04 at 11 56 40 AM

After you implement any fixes, you should make sure that `student_id=4168` still behaves properly.

Carifio24:
@patudom I think I found the issue with that 95% case and pushed a commit that I think solves it. Both of the cases we were looking at (student IDs 4094 and 4168) look good to me now; let me know what you think

PR #265, opened 2023-10-17
Calculate MC score based on wrong attempts rather than tries

### LINKED ISSUE #263: Update mc_radiogroup tries and scoring
New plan:

- Neutral responses should also increment ntries.
- Track wrong_tries and use that to calculate scores.

(See google doc that I sent on Slack for more details)

Carifio24:
This PR looks to resolve #263. To do this, the multiple choice component now tracks the number of wrong attempts in addition to the number of tries. Scoring is now done based on the number of wrong attempts, meaning that neutral attempts don't impact the score.

### DIFF
--- a/cosmicds/components/mc_radiogroup.vue
+++ b/cosmicds/components/mc_radiogroup.vue
@@ -76,7 +76,7 @@ module.exports = {
     points: {
       type: [Array, Function],
       default(_rawProps) {
-        return function(ntries) { return Math.max(12 - 2 * ntries, 0); };
+        return function(nwrong) { return Math.max(10 - 2 * nwrong, 0); };
       }
     },
     radioOptions: Array,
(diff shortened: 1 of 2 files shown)


patudom:
@Carifio24, sorry I didn't say this explicitly, but @johnarban and I felt that it looks weird on the dashboard for neutral questions to be awarded 0 points (because on a quick glance it can look like someone just failed to answer the question), so we think questions where all answers are neutral should also be awarded points.

I think that means that in this line https://github.com/cosmicds/cosmicds/blob/189f953d30b23afb69df642c4dc116dba30a461a/cosmicds/components/mc_radiogroup.vue#L133

we don't want `this.score` to be conditional based on `correct`, right?

cosmicds/components/mc_radiogroup.vue
```diff
@@ -112,29 +112,35 @@ module.exports = {
...
+      if (!forInitialization) {
         this.tries += 1;
+        if (wrong) {
+          this.wrongAttempts += 1;
+        }
       }
       this.complete = correct || (this.correctAnswers.length === 0 && neutral);
       if (this.scoring && this.complete) {
-        this.score = correct ? this.getScore(this.tries) : 0;
+        this.score = correct ? this.getScore(this.wrongAttempts) : 0;
```
johnarban, line 133: So I think to @patudom's comment, we would want it to be: 

```suggestion
        this.score = (correct || neutral) ? this.getScore(this.wrongAttempts) : 0;
```

patudom: @Carifio24, yes, if there is a correct answer and a neutral answer, we want the student to keep trying until they get to the correct answer. If all answers are neutral, any choice should give them 10 points. 

I don't recall if we have questions with some neutral and some wrong (but no correct) answers, but in that case, I would expect that any wrong choices would reduce their points, and choosing the neutral answer would provide them a score and allow them to advance.

PR #273, opened 2023-12-18
Add request logging

johnarban:
This PR adds logging of API requests from the python side by adding a [transport adapter](https://requests.readthedocs.io/en/latest/user/advanced.html?#transport-adapters) to the session that is created. 

It also logs the loading of the stages in `registries.py`

It updates the loading screen text for major steps in the `app.py` `_setup` script. The logs from the cosmicds app are labeled `Main App`. The hubbleds app creates two other sessions. 

The transport adapter has a random 3 digit identifier generated when it is created so that unique sessions can be easily identified

image

### DIFF
--- a/cosmicds/app.py
+++ b/cosmicds/app.py
@@ -1,6 +1,7 @@
 import json
 import os
 from os import getenv
+from requests.adapters import HTTPAdapter
 
 import ipyvuetify as v
 from cosmicds.utils import API_URL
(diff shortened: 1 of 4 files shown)


Carifio24:
Cool! It seems to be working well for me. This should help figuring out where we're running into issues communicating with the API server.

The only issue I see is that this adds a ton of blank space above the app in the notebook, which seem to result from the output of the `IPython.display.Javascript` calls. I'm not sure there's much that we can do about that, though.

patudom:
This is working pretty well for me too, but it always reports "None" for the student at the very beginning, even though it knows who the student is later. Does that seem ok to you? 

Regarding the blank spaces in the notebook, I'm assuming we can disable the logging code once we diagnose what the issue is on the server, so hopefully this won't be a long term issue.

Screen Shot 2024-01-03 at 4 24 49 PM

Carifio24:
@patudom The reporting the student username as `None` comes from [here](https://github.com/johnarban/cosmicds/blob/add-request-logging/cosmicds/app.py#L89). Since this is looking at the JupyterHub username environment variable, it makes sense that it's `None`. When we're on the JupyterHub this username should actually be populated.

patudom:
Ok, sounds good. Thanks, @Carifio24 and @johnarban! Merging now.

PR #275, opened 2024-04-30
Update mc radiogroup

johnarban:
This updates McRadioGroup to using a single $emit for the various `mc-` messages. It also moves initialization to mounted and adds an `initialization` prop so values can be set whenever it is displayed.


nmearl:
@johnarban Great update! We should start to minimize window-level events.

PR #277, opened 2024-05-03
CosmicDS viewer wrappers (with Plotly updates)

Carifio24:
This PR re-adds our custom CosmicDS viewer wrappers, with a few updates for the Plotly viewers. I've left the toolbar using the same setup since we had before, since it still works in the Solara version of the story.

Edit: This now also adds in the dotplot viewer and a `ViewerLayout` component.


nmearl (approved):
The `ViewerLayout` looks great. This all looks good to me.

PR #281, opened 2024-05-09
Add LayerToggle and LineFitTool

johnarban:
This PR adds the LayerToggle and LineFitTool needed in stage 6. [EDIT:}The LineFitTool has been update to work with plotly viewers


Carifio24:
Other than the issue with not updating, the layer toggle looks fine to me at the moment. I'm not sure why it doesn't update as it should attach the relevant callback to the viewer state inside its constructor, but maybe something with the current setup prevents that callback from triggering correctly.

I also feel like we should have the Vue template and the Python together in a folder inside of the `components` directory, something like below. It just feels strange to me to have the template and Python not be colocated. But maybe @nmearl has a different opinion?
```
components
+-- layer_toggle
|       LayerToggle.vue
|       layer_toggle.py

...other components
```

johnarban:
The LineFitTool has be updated to work with plotly viewers. 

@Carifio24 I am also not sure why they are split up. Right now I think it does load custom vue_components from  `vue_components`, which I believe is necessary for us to be able to use things like `<mc-radiogroup>` in `vue` files. 
but I think this would work regardless of how the directories are structured. 

I would be for combining these into 1

johnarban:
Currently, when used with [Stage 6](https://github.com/cosmicds/hubbleds/pull/364)

we get this error


```python
Traceback (most recent call last):
  File "/Users/jal194/anaconda3/lib/python3.9/site-packages/reacton/core.py", line 1661, in _render
    root_element = el.component.f(*el.args, **el.kwargs)
  File "/Applications/John/github/cosmicds_dev/hubbleds/src/hubbleds/pages/06-prodata/__init__.py", line 69, in Page
    component_state.add_data_by_marker(viewer)
  File "/Applications/John/github/cosmicds_dev/hubbleds/src/hubbleds/pages/06-prodata/component_state.py", line 113, in add_data_by_marker
    viewer.add_data(data)
  File "/Applications/John/github/cosmicds_dev/cosmicds/src/cosmicds/viewers/viewer.py", line 50, in add_data
    return super().add_data(data)
  File "/Users/jal194/anaconda3/lib/python3.9/site-packages/glue_jupyter/view.py", line 118, in add_data
    result = super().add_data(data)
  File "/Users/jal194/anaconda3/lib/python3.9/site-packages/glue/viewers/common/viewer.py", line 220, in add_data
    layer.update()
  File "/Users/jal194/anaconda3/lib/python3.9/site-packages/glue_plotly/viewers/scatter/layer_artist.py", line 254, in update
    self._update_display(force=True, **kwargs)
  File "/Users/jal194/anaconda3/lib/python3.9/site-packages/glue_plotly/viewers/scatter/layer_artist.py", line 166, in _update_display
    self._update_zorder()
  File "/Users/jal194/anaconda3/lib/python3.9/site-packages/glue_plotly/viewers/scatter/layer_artist.py", line 172, in _update_zorder
    self.view.figure.data = traces
  File "/Users/jal194/anaconda3/lib/python3.9/site-packages/plotly/basedatatypes.py", line 717, in __setattr__
    super(BaseFigure, self).__setattr__(prop, value)
  File "/Users/jal194/anaconda3/lib/python3.9/site-packages/plotly/basedatatypes.py", line 1001, in data
    raise ValueError(err_msg)
ValueError: The data property of a figure may only be assigned 
a list or tuple that contains a permutation of a subset of itself.
    Received duplicated traces
```

Carifio24:
I just pushed a couple of commits that, together with the new `glue-plotly`, have things almost working on my end. There's one more issue, which I suspect is an issue on this end, with layer toggle items not controlling the correct layers:

https://github.com/cosmicds/cosmicds/assets/14281631/81f5c187-1f3f-4499-851f-2a74e5844ee1

nmearl:
> @Carifio24 I am also not sure why they are split up. Right now I think it does load custom vue_components from vue_components, which I believe is necessary for us to be able to use things like <mc-radiogroup> in vue files.
but I think this would work regardless of how the directories are structured.

@Carifio24, @johnarban, these files are not co-located because they are not _technically_ related. Taking `ScaffoldAlert.vue`/`scaffold_alert.py` as an example:

- `scaffold_alert.py` does _not_ implement the `ScaffoldAlert.vue` in anyway. It basically wraps other `.vue` files (e.g. the guildelines), which themselves use `ScaffoldAlert.vue`. So the, in this instance, the distinction is that `ScaffoldAlert.vue` and `scaffold_alert.py` do not depend on each other, and are there separate.

The files in `vue_components` should only be `.vue` components; these get loaded into Solara and made available in other `.vue` files, but do not have a python component to them.

Since the layer tool is a python component fundamentally, it should follow the setup described by @Carifio24.

johnarban:
I have fixed the ordering of the layers in the layer toggle, but having it only respect the order of the layers in the viewer state.  

I also put the files in the `components` directory

johnarban:
The reason I am not using the zorder is that if both data sets are not added at the same time before the Layer Toggle is created. The newly added dataset gets `zorder = 0` while being later in the layer order.

Carifio24:
We discussed this on Slack, but just to leave this here for posterity:

So the zorder issue that we were seeing actually has its roots in the internals of glue - basically, when adding a layer, the layer state is added to `viewer.state.layers` (which triggers any callbacks on that property) as soon as it's created, but it's only added to the viewer's layer artist container (which is where zorder is initially set) after that. So during the callback execution the zorder won't have been set yet. To work around this, the solution that's now used is to instead hook into the layer artist container event instead so that the zorder has already been set (and the layer state will be inside of `viewer.state.layers`).

PR #287, opened 2024-06-03
Add dependencies

Carifio24:
While working on https://github.com/cosmicds/hubbleds/issues/380, I realized that it currently takes a bit of work to set up a development environment as we don't have any dependencies listed in the package file. This PR adds our current dependencies and slightly fleshes out the `pyproject.toml`.

FWIW, I determined the dependency list by grepping through the code for which (non-standard-library) packages we import from.


Carifio24:
I realized after doing this that we have both `pyproject.toml` and `setup.cfg`. @nmearl do you have any preference on which of these we use? I would assume we want `pyproject.toml` but don't have strong feelings either way

PR #293, opened 2024-06-15
Add statistics and percentage selector components

Carifio24:
This PR ports the statistics and percentage selectors over to Solara. The logic of these was mostly Python-side anyway, so I decided to just go ahead and move the layout side to Solara as well. Additionally, this adds a utility function based on what @nmearl did for the dotplot viewer to encapsulate the operations for making a Plotly figure resizable.

These mostly behave the same way as they did previously, but any thoughts or tweaks to the layout are welcome.

### DIFF
--- a/src/cosmicds/components/__init__.py
+++ b/src/cosmicds/components/__init__.py
@@ -4,3 +4,4 @@ from .plotly_support.plotly_support import PlotlySupport
 from .viewer_layout import *
 from .debug_control import StateEditor
 from .layer_toggle import LayerToggle
+from .statistics_selector import StatisticsSelector
--- a/src/cosmicds/components/__init__.py
+++ b/src/cosmicds/components/__init__.py
@@ -5,3 +5,4 @@ from .viewer_layout import *
 from .debug_control import StateEditor
 from .layer_toggle import LayerToggle
 from .statistics_selector import StatisticsSelector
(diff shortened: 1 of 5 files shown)


patudom:
Thanks, @Carifio24! This is looking good. I do have some comments about the component styling, but let's leave cosmetic changes until after we've worked out all the content. Are the data points all just integers? I'm wondering why the histogram bars are not centered on their respective values, whereas the mode line markers are. This will likely confuse a lot of users, so we should get the histogram bars centered if we can.

Screenshot 2024-06-17 at 10 39 42 AM

Carifio24:
@patudom So I think the issue that we're having here is that the histogram bins aren't playing nicely with the integer values of the points. In the histogram viewer state, you set the number of bins. This doesn't play particularly nicely with our use case, since the students don't have any way to adjust the bins themselves.

It looks like what we did in the past was to [explicitly set the number of bins to be equal to the range of the data](https://github.com/cosmicds/hubbleds/blob/legacy/src/hubbleds/stages/stage_5.py#L760) - basically, make the bin edges be the integer values within the range. I'll implement the same thing in the new version (that'll go over in Hubble).

Carifio24:
@patudom I've added that bin-updating functionality to the Hubble PR.

PR #298, opened 2024-06-26
Add piggybank_total to global state & style chip

patudom:
Marking this as a draft because I think @nmearl said he would be changing where the points total will live after the GLOBAL/LOCAL/Component state refactoring.


patudom:
I merged Nick's revamped state content to here, so this should be current. All it really does now is it restyles the piggybank chip to look how it looked before. It is not correctly wired up to show the updating points.

PR #300, opened 2024-07-01
Update state editor for new state structure

Carifio24:
This PR updates the state editor and associated components to work with the new state structure.

Builds on #299.


patudom:
Looks like this got merged to @nmearl's main, but not here, so I've reopened this to get these 2 commits, else what's on hubble will break.

PR #303, opened 2024-07-08
Updates to handle student data

Carifio24:
This PR contains some various updates to the base package. These are issues that popped up while I was working on using the real student data in stages 4 + 5 of the Hubble app, but should be generally useful changes. The main changes include:

* Remove the subset selection tools from our base CosmicDS viewers, since we tend to not give students the ability to make selections by default
* Allow our JSON encoder to serialize `datetime` objects
* Add a method to the global state to allow either adding data (if it's not present) or updating existing data, based on the data's label. Using this method prevents accidentally adding the same data to the data collection twice
* Add a utility function for creating an empty glue `Data` object with appropriately-typed components from a `pydantic` model class

### DIFF
--- a/src/cosmicds/layout.py
+++ b/src/cosmicds/layout.py
@@ -5,7 +5,7 @@ from importlib.metadata import version
 
 import solara
 from solara.alias import rv
-from solara.lab import theme as theme, Ref
+from solara.lab import theme, Ref
 from solara.server import settings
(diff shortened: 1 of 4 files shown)


nmearl (approved):
I really like the use of the Pydantic field type annotations to determine the Glue component class. 👍

PR #304, opened 2024-07-09
Add button for refreshing stage state and refactor API functionality

Carifio24:
This PR allows us to reset the stage state for the current student/story/stage. To allow this, I've done the following:
* Create a "refresh button" component that reloads the current page and allows running some code pre-refresh. While Solara has a router that we can access, setting the path to the current URL has no effect, so this was the simplest way that I could think of to accomplish this
* Add a "reset stage state" button to the state editor component that uses the refresh button component and deletes the current stage state from the database
* Create a `BaseLocalState` class with a few common story state fields, so that we can reference them here
* To go along with this, I've moved the generic stage/story state manipulation endpoints here to the base package. I decided not to implement the `PUT` methods here, as these will generally involve things like ignoring fields that we can leave to a specific story.

### DIFF
--- a/src/cosmicds/components/__init__.py
+++ b/src/cosmicds/components/__init__.py
@@ -6,3 +6,4 @@ from .debug_control import StateEditor
 from .layer_toggle import LayerToggle
 from .statistics_selector import StatisticsSelector
 from .percentage_selector import PercentageSelector
+from .refresh_button import RefreshButton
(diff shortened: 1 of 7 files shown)


patudom:
@johnarban, FYI, if you are using the `StateEditor` in stage 3, be aware that you now have to also pass the local state & local api when you call it:
```
    StateEditor(Marker, COMPONENT_STATE, LOCAL_STATE, LOCAL_API)
```

PR #308, opened 2024-07-10
Fix typo in info panel

Carifio24:
This is a tiny PR to fix a typo that was bothering me.


patudom:
Those usually bother me too, but I didn't even notice this one. 😄

PR #311, opened 2024-07-16
update styling for various elements

patudom:
This plus https://github.com/cosmicds/hubbleds/pull/488 is my attempt to bring the layout more inline with what we have in the voila version.

I've also made some minor adjustments to the guideline, MC, and FR components so there is sufficient contrast btw text and background in both light and dark modes.


patudom:
Merging since this is just cosmetic and it will be easier to see where there are problems as we test other things.

PR #322, opened 2024-07-30
Add gaps between histogram bars by default

Carifio24:
This PR updates the defaults for our histogram viewer states to resolve https://github.com/cosmicds/hubbleds/issues/556. I assume we'll generally want gaps between bars, so I made it the default here rather than at the Hubble level.

@patudom Let me know if we want to adjust the gap sizing - the gap fraction indicates what percentage of the plot's horizontal space is taken up by the gaps (so `0.5` would mean that the bars and gaps are equally sized, for example).


patudom:
This looks great - thanks, Jon!

PR #336, opened 2024-09-18
Use generic to avoid linting errors with base states.

Carifio24:
This PR gets rid of the annoying linting errors where the linter thinks that Reactive[T] isn't a match for Reactive[S] when T is a subclass of S.


Carifio24:
Note that I only did this for the state editor, but we could use the same pattern elsewhere. Also, I marked these as covariant since, in the case of the state editor, we're using `Reactive` as a producing interface (via the `.value` method), but aren't using the consuming portion of it (i.e. `.set`. We set field refs, but don't reassign the state instance itself). In general `Reactive` is invariant, though.

PR #337, opened 2024-09-23
Update student creation request

Carifio24:
This PR updates the form of the student signup POST request body. We updated the schema of this endpoint while working on the portal and thus needs to be updated here.

I tested this locally with our usual and my new student was added to it.


Carifio24:
Note that this could be related to https://github.com/cosmicds/hubbleds/issues/618.

PR #338, opened 2024-09-25
Fix issues with speech autochange

Carifio24:
This PR should resolve https://github.com/cosmicds/hubbleds/issues/613. With these changes, the behavior for the slideshows should be as follows:

* If autoread is set, changing slides will cause the current speech to stop, and the next slide will automatically start to be read
* If autoread isn't set, changing slides will still cause the current speech to stop

### DIFF
--- a/src/cosmicds/components/speech_settings/SpeechSettings.vue
+++ b/src/cosmicds/components/speech_settings/SpeechSettings.vue
@@ -79,21 +79,13 @@
 
 <script>
 module.exports = {
-  props: {
-    initialState: {
-      type: Object,
-      default: null,
-    }
-  },
   created() {
     window.speechSynthesis.onvoiceschanged = (_event) => {
       this.updateVoiceList();
     };
     this.updateVoiceList();
 
-    this.autoread = this.initialState?.autoread ?? false;
-    this.pitch = this.initialState?.pitch ?? 1;
-    this.rate = this.initialState?.rate ?? 1;
+    this.updateAttributes(this.initialState);
   },
   data() {
     return {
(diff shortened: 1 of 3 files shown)


patudom:
Sorry if these issues are because I forgot about this PR and it has languished for a while.😬 

This doesn't fix the slideshow issues for me. It will still keep reading the first slide when I move on to other slides. (I also have to press the play/pause button a couple times before it starts playing).

Turning on the autoplay setting gives these errors in the terminal:
```
ERROR:    Uncaught exception: Traceback (most recent call last):
  File "/Users/Pat/anaconda3/envs/common-solara/lib/python3.11/site-packages/ipywidgets/widgets/widget.py", line 191, in __call__
    local_value = callback(*args, **kwargs)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/Pat/anaconda3/envs/common-solara/lib/python3.11/site-packages/ipyvue/VueTemplateWidget.py", line 60, in _handle_event
    getattr(self, "vue_" + event)(data)
  File "/Users/Pat/anaconda3/envs/common-solara/lib/python3.11/site-packages/solara/components/component_vue.py", line 29, in event_handler
    callback(data)
  File "/Users/Pat/anaconda3/envs/common-solara/lib/python3.11/site-packages/cosmicds/layout.py", line 190, in <lambda>
    event_autoread_changed=lambda read: update_speech_property("read", read),
                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/Pat/anaconda3/envs/common-solara/lib/python3.11/site-packages/cosmicds/layout.py", line 186, in update_speech_property
    setattr(settings, prop, value)
  File "/Users/Pat/anaconda3/envs/common-solara/lib/python3.11/site-packages/pydantic/main.py", line 795, in __setattr__
    raise ValueError(f'"{self.__class__.__name__}" object has no field "{name}"')
ValueError: "Speech" object has no field "read"
```

Carifio24:
@patudom Just curious, are you on the right branch? Changing `"read"`-> `"autoread"` in that layout call is one of the changes in this PR.

patudom:
Oh weird - I was on the correct branch, but it was behaving like I hadn't installed it in editable mode. I reinstalled and it's working now.

I am still having the problem where the first time I click the talking icon, nothing happens, and I have to click it again. After that first time, it behaves normally.

https://github.com/user-attachments/assets/49418897-5833-435c-adee-752ade761e7e

Carifio24:
This can happen for me as well, though it only seems to happen sometimes.

PR #339, opened 2024-10-03
Minor fix to Contrast Brightness Control

johnarban:
The control needs to emit the new brightness/contrast values when it is reset. We need this if we want to track the brightness like we did in the voila version.


patudom:
It doesn't look like my brightness got stored in the db when I tested. I will try again and report back.:
Screenshot 2024-10-04 at 6 41 41 PM

PR #342, opened 2024-10-04
Update global and local state without changing instance

### LINKED ISSUE #341: Modify how we update global state to not create new instance
Currently we update the instance stored in the global state reactive when we fetch the state from the database (see [here](https://github.com/cosmicds/cosmicds/blob/main/src/cosmicds/remote.py#L202)). We should change this so that we just mutate the underlying instance. If there are properties that we need to update reactively, we can always do that explicitly with refs.

Carifio24:
This PR is intended to fix #341. The motivation for this is @johnarban's first note on https://github.com/cosmicds/hubbleds/pull/629. The issue is that we're grabbing the initial glue application information with the original global state instance, but then when we reset the global state instance that has a new data collection, and we then have two different ones floating around.

The approach here is to update via a dictionary (which I imagine will always be the JSON content of a request response). We iterate through the dictionary items and, if the key corresponds to an attribute of the state, `Ref`-update the state accordingly. The update method is simple and assumes that the only special handling needed is for Pydantic models.

In the Hubble story we only need to call this once (when the state-loading task finishes) and probably only need reactivity for a subset of fields, but I didn't want this approach to make any assumptions about what we'll generally want.

### DIFF
--- a/src/cosmicds/remote.py
+++ b/src/cosmicds/remote.py
@@ -1,4 +1,3 @@
-from pydantic import BaseModel
 from solara_enterprise import auth
 import hashlib
 import os
@@ -225,14 +224,8 @@ class BaseAPI:
 
     @staticmethod
     def _update_state(state: Reactive[BaseState], data: dict):
-        for key, value in data.items():
-            field = getattr(state.fields, key, None)
-            if field is None:
-                continue
-            state_value = getattr(state.value, key)
-            if isinstance(state_value, BaseModel):
-                value = state_value.__class__(**value)
-            Ref(field).set(value)
+        new_state = state.value.__class__(**data)
+        state.value.__dict__.update(new_state.__dict__)
 
 
 BASE_API = BaseAPI()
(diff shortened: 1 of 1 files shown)


Carifio24:
@nmearl I've updated this to only update the model instance wrapped by the reactive. It seems that [the best way to do this is by updating the model's `__dict__`](https://stackoverflow.com/questions/73421846/how-to-update-pydantic-model-from-dictionary). I'm really surprised that there isn't some sort of `update_from(other: MyModel)` style function, but oh well.

To deal with the fact that we often have nested models, I just took the easy way out and created a new instance, then used its `__dict__` to update the original model instance. I started writing my own function to try and handle reformatting the dictionary and parsing sub-models, but there are a lot of cases to consider, many of which will need some recursive calls, so I figured I'd just let Pydantic do it rather than write some complicated thing that we then need to maintain.

nmearl:
Sounds good. And just to confirm, this keeps the reference to the same glue instance?

Carifio24:
Yeah, the cached properties aren't included in the `__dict__` so the glue items (data collection, session, etc.) aren't changed.

nmearl (approved):

PR #344, opened 2024-10-15
Create Google Analytics support component

Carifio24:
This PR adds a Google Analytics support component, similar to what we've done for MathJax and Plotly. This component takes in the tracking tag to be used so that we can use different tags for different stories if necessary.


patudom:
merging this in so we can start testing it on a deployed branch.

PR #345, opened 2024-10-17
Update demo to not read states

johnarban:
This PR fixes https://github.com/cosmicds/hubbleds/issues/659, preventing the reading and global, local, and component states from the database. This is a companion to https://github.com/cosmicds/hubbleds/pull/665, and https://github.com/cosmicds/hubbleds/pull/671

### DIFF
--- a/src/cosmicds/layout.py
+++ b/src/cosmicds/layout.py
@@ -106,7 +106,7 @@ def BaseLayout(
         with rv.AppBar(elevate_on_scroll=False, app=True, flat=True, class_="cosmicds-appbar"):
             
             rv.Html(tag="h2", children=[f"{story_title}"])
-            rv.Html(tag="h3", children=['Cosmic Data Stories'], class_="ml-4")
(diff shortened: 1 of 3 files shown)


patudom:
I updated this to merge to the public-demo branch instead of main. Is it helpful to also have these on main?

patudom:
Per team call, the plan is to not use a demo branch, so we don't have to maintain 2 separate branches. Instead, we will set up an environment variable that determines whether we turn the retrieval off/on.

Carifio24 (approved):

patudom:
@johnarban and @Carifio24, should we set up the environment variable before we merge this to main?

Carifio24:
I actually think the piece to read an environment variable needs to be added (unless I missed it somewhere).

johnarban:
This has been merged with main, and now uses the ` CDS_DISABLE_DB` variable to set the value of GlobalState.update_db, by checking that the environment variable exists and has a value of `true`. @Carifio24 @nmearl

PR #349, opened 2024-10-29
Use `expected_size` field to set class size in app

Carifio24:
This PR updates the app to use the teacher-set "expected size" for the class, rather than the number of students currently in the class.

### DIFF
--- a/src/cosmicds/remote.py
+++ b/src/cosmicds/remote.py
@@ -54,9 +54,9 @@ class BaseAPI:
     def update_class_size(self, state: Reactive[GlobalState]):
         class_id = state.value.classroom.class_info["id"]
         size_json = self.request_session.get(
-            f"{self.API_URL}/classes/size/{class_id}"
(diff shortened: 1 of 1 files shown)


patudom:
@Carifio24, do we need to have some sort of fallback if the teacher did not enter an expected class size?

Carifio24:
@patudom We probably do want a fallback, yeah. It could be the "actual" class size (that is, how many students are associated with the class in the database).

patudom:
Yeah, that makes sense - can we do what we would have done before, which I think was to count the class size at the beginning of Stage 5?

patudom:
@Carifio24 & I just discussed this - we don't think that we use the class size anywhere as a state variable. (We do keep track of class size on the db side). Jon will confirm - if that's the case, we can close this PR.

Carifio24:
Yeah, I don't think we're actually using this at all - we set this value in a couple of places, but we never read it. We can reopen if we ever do need to update how we determine the size.

PR #351, opened 2024-11-21
Make components more generic

Carifio24:
While working on setting up the Solara app/layout for the TEMPO story, I came across a couple of Hubble-specific things in our base components:

* The base layout expects `hubbleds` to be installed, as it shows the Hubble version in the debugging info. I think this is useful while we're still working on it, but I've wrapped the component in a try-catch so that it can still run if `hubbleds` isn't present in the environment
* The `ViewerLayout` component sets up some auto-resizing functionality for the Plotly viewers, but this means that it doesn't work to wrap a non-Plotly viewer. This PR fixes that.


nmearl (approved):
Nice fixes!

PR #361, opened 2025-01-08
Use "default" states when not loading from database

Carifio24:
This is an attempt to fix https://github.com/cosmicds/hubbleds/issues/741. This updates the remote handler to explicitly set the current state to an empty "default" state when we aren't reading from the database.

@patudom @johnarban if you guys could try this out and tell me whether it works for you as well, that would be great


patudom:
I don't know if this is due to this branch or if this is an unrelated problem, but I'm having an issue on safari.
- on the main hubbleds branch
- for a fresh student not in the db
- logging in via a google authentication.

What happens is I pass the authentication, the introduction loads, but then it boots me back out to the login screen. This does not happen on Chrome.

If the student already exists, Safari loads properly with no issues.

PR #362, opened 2025-01-12
Update for anonymized id tokens

nmearl:
This PR handles the new anonymized user info format retrieved from auth0.


johnarban:
This error occurs in the `BaseLayout` because the format of the returned information is different. `name` and `email` are no longer the correct keys, but `cds/name` and `cds/email`. 
image

This commit fixes that 
https://github.com/cosmicds/cosmicds/pull/362/commits/00856c7d53395612dfeb97a64d48b94b461bf95a

nmearl:
Thanks for catching the missed layout user info. We already have fall back handling wrapped up in the `display_info` computed property, so I've added some changes to rely on that explicitly.

PR #371, opened 2025-03-13
Keep local setting from overriding dark theme default

patudom:
Per our meeting today, explicitly set dark theme so the ThemeToggle component doesn't override with the user's system setting. This will close https://github.com/cosmicds/hubbleds/issues/806


johnarban (approved):
this looks good to me. works locally

PR #375, opened 2025-03-25
Implement local theme toggle with improved handling

johnarban:
This PR creates our own version of Solara's `ThemeToggle`, which adds the ability to manually set a default theme and  use the theme defined by the server via `--theme-variant`. Like the original version, the toggle always respects the user's local storage settings when loading and never ignores them. 

Two main examples for how to use this are:

- Allow "auto" theme detectino but intialize with a default theme. 
```python
ThemeToggle(
    on_icon="mdi-weather-night",  # dark mode icon
    off_icon="mdi-brightness-7",  # light mode icon
    enable_auto=True,             # allow auto theme detection
    default_theme="dark",         # default to dark theme
    enforce_default=True,         # enforce the default theme
)
```

Same as above, but use the server-defined theme as the default
```python
ThemeToggle(
    on_icon="mdi-weather-night",  # dark mode icon
    off_icon="mdi-brightness-7",  # light mode icon
    enable_auto=True,             # allow auto theme detection
    default_to_server=True,       # use server-defined theme
    enforce_default=True,         # enforce the default theme
)
```

Simple setup to respect the server default and don't allow the user to set the 'auto' setting
```python
ThemeToggle(
    on_icon="mdi-weather-night",  # dark mode icon
    off_icon="mdi-brightness-7",  # light mode icon
    enable_auto=False,             # allow auto theme detection
    default_to_server=True,       # use server-defined theme
)
```

### Explanation of new options

- `enable_auto`: This works the same as before. If set to False and no default is configured, it will use the user's `preferred-color-scheme` settings.
- `default_to_server`: Previously the toggle ignored the server setting. If `True`, and either `enable_auto = False` or `enforce_default = True`, the app theme will respect the server settings
- `enforce_default`: Forces the initial theme of the app to be the default theme (or the server theme) specified.

A test repository can be found [here](https://github.com/johnarban/solara-theme-toggle-test)

### DIFF
--- a/src/cosmicds/components/theme_toggle/__init__.py
+++ b/src/cosmicds/components/theme_toggle/__init__.py
@@ -1,0 +1,1 @@
+from .theming import ThemeToggle
--- a/src/cosmicds/components/theme_toggle/theming.py
+++ b/src/cosmicds/components/theme_toggle/theming.py
@@ -1,0 +1,98 @@
+from typing import Callable, Dict, Union, cast
+
+import ipyvuetify.Themes
+from ipyvuetify.Themes import Theme
+
+import solara
+from solara.components.component_vue import component_vue
+from solara.tasks import Proxy
+
+theme = Proxy(Theme)
+ipyvuetify.Themes.theme = cast(ipyvuetify.Themes.Theme, theme)
+
+
+def use_dark_effective():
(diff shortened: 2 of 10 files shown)


patudom:
This is behaving well for me locally - thank you, @johnarban! Should we remove the console logs?

PR #384, opened 2025-04-01
Base layout changes

nmearl:
With the recent changes to the display user info and the navigation drawer, things were getting a tad misaligned and a bit spread out. I've made some minor changes that I think help to clean up the story chrome:

- No longer clip the App bar and have it extend across the top. Use the built-in navigation drawer toggle icon.
- Implement a `BreakpointWatcher` that notifies Solara when Vuetify has triggered a breakpoint change. This allows us to know when the navigation drawer has switched to floating. Doing so allows us to render a small "close" button only when the navigation drawer is floating.
- Revert the list back to the default formatting. I believe the changes were due to some stages escaping their `div`s, but we can fix that by adjusting the margins around the list item icons.
- Centralize the user info into a shelf that slides up from the bottom of the navigation drawer.
- Consequently, re-order the user score and user settings and visually distinguish their locations with a vertical divider

https://github.com/user-attachments/assets/4adf88c7-4efd-4a81-bf30-d256d63e794e

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -57,9 +57,8 @@ install_requires =
     ipyvuetify
     numpy<2.0.0
     reacton
-    solara==1.42.0
-    solara-enterprise==1.42.0
-    solara-ui @ git+https://github.com/nmearl/solara.git@v1.42
+    solara==1.44.1
+    solara-enterprise==1.44.1
     traitlets
 
 [options.packages.find]
--- a/src/cosmicds/components/breakpoint_watcher/BreakpointWatcher.vue
+++ b/src/cosmicds/components/breakpoint_watcher/BreakpointWatcher.vue
@@ -1,0 +1,39 @@
+<script>
+export default {
+  name: "BreakpointWatcher",
+  data() {
+    return {
+      previousBreakpoint: null,
+      currentBreakpoint: null,
+    }
+  },
+  mounted() {
+    window.addEventListener("resize", this.onResize);
+    this.onResize();
+  },
+  unmounted() {
+    window.removeEventListener("resize", this.onResize);
+  },
+  methods: {
+    onResize() {
+      const h = window.innerHeight
+      const w = window.innerWidth
+      const bp = this.$vuetify.breakpoint.name
+
+      if (this.previousBreakpoint !== bp) {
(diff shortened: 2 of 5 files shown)


patudom:
This is great, @nmearl! I really like how you handled the top title bar going all the way across the full width of the page and putting the menu under it. Wish I'd thought of that.😅 

I have few couple comments about the student info:
- Within a classroom context, it will be easier for teachers if their students don't have to click an extra button to figure out what their ID is (especially if a teacher hasn't made use of the dashboard feature to upload a local csv file that translates student IDs into student names). I would prefer for the ID # to be visible at the bottom of the menu (in place of "Student Info").
- We promise teachers and district IT folks that we are not collecting private information on students. Even though we are hashing the emails in our db, if the interface can still display their email addresses (which I'm assuming are stored locally), most general users will not understand that we are not storing this information, so we should remove the student email address from the display. (This is why I removed it from the student account interface on the portal).
- I'm not sure it's useful for students to know what their hashed ID is. Can others think of a reason? If there isn't one, I'm inclined to get rid of the pop up menu all together and just display the student ID.

Let me know what you think! If this sounds ok to you, I can implement these changes.

patudom:
One more minor thing. I think of the light/dark mode and speech settings as being something you set once and then ignore for the rest of the experience, whereas the points total continually changes. Because of this, I think the points should have the more prominent position and be switched back to the upper right corner.

nmearl:
> I'm not sure it's useful for students to know what their hashed ID is. Can others think of a reason? If there isn't one, I'm inclined to get rid of the pop up menu all together and just display the student ID.

The educator portal displays the students in a class by their hashed ID, so for an educator to remove a particular student, they'd have to know their hashed ID. 

More generally, I've also been told to never expose the index ID of your database to the public (what we've been calling the "Student ID"). I definitely do not have a ton of experience with managing large scale databases, but from what I've been told, exposing your index ID opens you up to enumeration attacks because auto-incrementing IDs are predictable. Likewise, exposing sequential IDs can leak information about how many users exist. There are some other things like issues with database merging, migrating, etc, where IDs can be re-arranged and therefore no longer reference the row you thought it did, but I don't anticipate we'll be doing that.

Otherwise, I see no issues with your changes, and I can make them pretty quick.

PR #385, opened 2025-04-01
Show Educator Demo Interface

johnarban:
Used the hashed identifier to determine if the user logging in is an educator, and if so show them the team interface. There is a companion PR on hubblds https://github.com/cosmicds/hubbleds/pull/960

Once it determines the user is an educator, it uses the demo student. In the database, I have duplicated the demo student as an educator so that when it checks again it will still see the user as an educator.

This roundabout was needed because the `GlobalState` can get reset or reinitialized, and will reset to default values (if they're not stored in the DB). Since I am trying to follow https://github.com/cosmicds/hubbleds/issues/951, we couldn't do that. Also, url params are not able to be used since the url gets cleared.

### DIFF
--- a/src/cosmicds/layout.py
+++ b/src/cosmicds/layout.py
@@ -69,6 +69,7 @@ def BaseSetup(
     educator_mode = False
     if bool(auth.user.value):
         if BASE_API.is_educator:
+            force_demo = True
             educator_mode = True
             GLOBAL_STATE.value.update_db = False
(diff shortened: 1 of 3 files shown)


patudom:
For context, @nmearl, the goal here is to add a "launch" button to the educator portal that teachers can use to run the full non-demo version of the story with these parameters: https://github.com/cosmicds/hubbleds/issues/951

PR #386, opened 2025-04-02
Use memo on initial page state load

nmearl:
Explicitly define the `dependencies` to be empty in hooks to avoid potentially weird behavior (according to Solara docs). See cosmicds/hubbleds#964 for more information.

### DIFF
--- a/src/cosmicds/layout.py
+++ b/src/cosmicds/layout.py
@@ -85,22 +85,25 @@ def BaseSetup(
         )
         class_code.set("215")
 
-    if bool(auth.user.value):
-        if BASE_API.user_exists:
-            BASE_API.load_user_info(story_name, GLOBAL_STATE)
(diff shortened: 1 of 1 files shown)


nmearl:
This now has ramifications for the Demo -- it will need to be undated.

patudom:
Thanks, Nick. Once we finish testing them, I'm wondering if we can replace the short demo branch with @johnarban's #960 and https://github.com/cosmicds/cosmicds/pull/385. The main motivation behind creating the short demo version back when we did is that it let us skip a bunch of the content that had not been fully debugged in time for teacher workshops. 

If that works out, we can discuss whether it makes sense to fully retire the demo server and redirect the demo url to the educator mode version. Whatever we do, we'll still want newcomers to be able to trial the educator mode without needing a login.

PR #387, opened 2025-04-04
Adjust DotPlot Point size

johnarban:
Adjust the `dotplot` point radius by using our own `CDSPlotlyDotplotLayerArtist`.


patudom:
Thank you, @johnarban! This is definitely an improvement, but it seems like it is not being invoked when the viewer is first drawn (bottom). If I zoom in then zoom back out, the dots have the larger size (top). 
Screenshot 2025-04-04 at 4 56 26 PM

patudom:
This is being fixed directly in https://github.com/glue-viz/glue-plotly/pull/124

PR #388, opened 2025-04-05
Handle user authentication state

nmearl:
This PR adds in functionality to attempt resolving user authentication state, and on failure, force the user back to the portal app. Note that this will not quite work when running locally, since it's not possible to parse the portal url from a story app due to the differing ports (it should route you back to the server base url). On the deployed server, however, this should take the user back to the portal.

Also of note is that the auth reactive variable is unique to each server, so it is not really possible to determine when a user logs out from a different location unless they refresh the entire page. However, this may be different on the deployed server in which the url root is the same. Therefore, I'd like to check this after deploying, and if it's not the case that the auth state is propagated along url root paths, then we'll have to think of a solution to checking the auth state in the stories themselves.

### DIFF
--- a/src/cosmicds/components/location_helper/LocationHelper.vue
+++ b/src/cosmicds/components/location_helper/LocationHelper.vue
@@ -1,0 +1,12 @@
+<script>
+export default {
+  name: "LocationHelper",
+  created() {
+    window.location.href = this.url;
+  },
+}
+</script>
+
+<template>
+
+</template>
--- a/src/cosmicds/components/location_helper/location_helper.py
+++ b/src/cosmicds/components/location_helper/location_helper.py
@@ -1,0 +1,14 @@
+import solara
+
+
+@solara.component_vue("LocationHelper.vue")
+def LocationHelper(url: str):
+    """
+    Forcibly sets the window location to the given url on instantiation.
+
+    Parameters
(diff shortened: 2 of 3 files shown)


johnarban:
This seemed to work as described. Did two tests (with `main` merged in):

1. Logged in through the portal, opened the Hubble app, logged out from the portal, and refreshed the Hubble app. 
   - Result: After refresh, the Hubble app redirected to its base URL.
2. Logged in via the portal, shutdown `cds_portal` solara server, opened the Hubble app on port 8865, and clicked logout.
   - Result: The app redirected to the base URL.

nmearl:
@johnarban Can you try one more scenario: when logged into both the portal and Hubble app, log out from the portal, and then attempt to navigate to another page in the Hubble app?

johnarban:
So after logging out from the portal, I can continue using the app as normal, progressing through the story and flipping back between stages. Only after I refresh the page or manually change the url does it realize it is not logged in, and then pushes itself to the base url

patudom:
Do we want to merge this with the behavior @johnarban described above, or does this still need updating?

nmearl:
I'd prefer to merge this and work off it in a future PR. I'm interested to see if the authentication state is split by base url.

